PR #130, opened 2021-03-13
Implemented Cheatin' Varmint

AaronSchnabel:
Cheatin' Varmint works, can choose either the regular Resolution or Cheatin' Resolution and it passed all tests on my end.


server/game/cards/01-DTR/CheatinVarmint.js
```diff
@@ -0,0 +1,31 @@
+const ActionCard = require('../../actioncard.js');
+
+class CheatinVarmint extends ActionCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Resolution: Lower hand by 2 ranks',
+            playType: ['resolution'],
+            choosePlayer: true,
+            cost: ability.costs.payGhostRock(5),
+            message: context =>
```
mmeldo, line 10: The hand rank modification is exception from the rule that we should use `message` property. That is because message property writes before the rank is manipulated, so the "Current rank is {}" will show rank before it is manipulated. 
Therefore cards doing card manipulations should write messages to log in `handler` after the manipulation.

server/game/cards/01-DTR/CheatinVarmint.js
```diff
@@ -0,0 +1,31 @@
...
+                this.game.addMessage('{0} pays 5 ghost rock and uses {1} to lower {2}\'s draw hand by 2 ranks. Current rank is {3}',
```
mmeldo, line 11:
```suggestion
                this.game.addMessage('{0} uses {1} and pays 5 ghost rock to lower {2}\'s draw hand by 2 ranks; Current rank is {3}', 
```

Just to adhere to the format {player} uses {source} to do something.

server/game/cards/01-DTR/CheatinVarmint.js
```diff
@@ -0,0 +1,31 @@
+const ActionCard = require('../../actioncard.js');
+
+class CheatinVarmint extends ActionCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Resolution: Lower hand by 2 ranks',
+            playType: ['resolution'],
+            choosePlayer: true,
+            cost: ability.costs.payGhostRock(5),
```
mmeldo, line 9: There is another cost type `payReduceableGRCost(amount)`, but I believe you used the correct one as it is not really a cost of the card. Or do you think this cost can be reduced? Is there actually a card that can reduce costs of action cards?

mmeldo (changes requested):

server/game/cards/01-DTR/CheatinVarmint.js
AaronSchnabel: I do not think there is another card that can lower the costs of an ability on a card rather than the cost of the card itself, though I could be wrong.

server/game/cards/01-DTR/CheatinVarmint.js
AaronSchnabel: Ah, that makes sense. I should have caught that.

server/game/cards/01-DTR/CheatinVarmint.js
mmeldo: Ok. I guess we will leave it like this. If there is a problem with it, we can fix later.

server/game/cards/01-DTR/CheatinVarmint.js
```diff
@@ -0,0 +1,31 @@
...
+            message: context =>
+                this.game.addMessage('{0} uses {1} and pays 5 ghost rock to lower {2}\'s draw hand by 2 ranks; Current rank is {3}', 
+                    context.player, this, context.chosenPlayer, context.chosenPlayer.getTotalRank()),
+            handler: context =>
+                context.chosenPlayer.modifyRank(-2)
```
mmeldo, line 14:
```suggestion
            handler: context =>
                context.chosenPlayer.modifyRank(-2);
                this.game.addMessage('{0} uses {1} and pays 5 ghost rock to lower {2}\'s draw hand by 2 ranks; Current rank is {3}', 
                    context.player, this, context.chosenPlayer, context.chosenPlayer.getTotalRank());       
```

server/game/cards/01-DTR/CheatinVarmint.js
```diff
@@ -0,0 +1,31 @@
...
+        });
+        this.action({
+            title: 'Cheatin\' Resolution: Lower hand by 2 ranks',
+            playType: ['cheatin resolution'],
+            choosePlayer: true,
+            message: context =>
+                this.game.addMessage('{0} uses {1} to lower {2}\'s draw hand by 2 ranks. Current rank is {3}', 
+                    context.player, this, context.chosenPlayer, context.chosenPlayer.getTotalRank()),
+            handler: context =>
+                context.chosenPlayer.modifyRank(-2)
```
mmeldo, line 24:
```suggestion
            handler: context =>
                context.chosenPlayer.modifyRank(-2);
                this.game.addMessage('{0} uses {1} and pays 5 ghost rock to lower {2}\'s draw hand by 2 ranks; Current rank is {3}', 
                    context.player, this, context.chosenPlayer, context.chosenPlayer.getTotalRank());       
```

Here are the suggested changes to move the message.

mmeldo (changes requested):

AaronSchnabel:
Committed your changes, but now it's failing. I'll take a look at it tomorrow and figure out what's wrong.

mmeldo:
Yeah, I did not check the lint so did not realize I put the semicolon after the modifyRank function which link did not like.

mmeldo (approved):

PR #185, opened 2021-03-19
Improve callout messaging

### LINKED ISSUE #186: Improve messaging for Callout action

mmeldo:
Closes #186


server/game/GameActions/CallOut.js
```diff
@@ -14,7 +14,7 @@ class CallOut extends GameAction {
 
     createEvent({ caller, callee, isCardEffect = true, canReject = true }) {
         return this.event('onCardCallOut', { caller: caller, callee: callee, isCardEffect: isCardEffect, canReject: canReject }, event => {
-            event.caller.callOut(event.callee, event.canReject);
+            event.caller.callOut(event.callee, event.canReject, event.isCardEffect);
```
tybar, line 17: Should there be a unit test that handles isCardEffect?

Though looking at this, there's nothing inherent about isCardEffect...is that going to be implemented in the future?

If so, then I think we can wait until that is implemented to write that test.

mmeldo: This PR is not yet completed. Usually, when I have PR completed I request a review. We can discuss how we can do it. For example in work, we have a special label that says the PR is ready for review.

tybar: That's probably a good way to do that honestly.

tybar (approved):

mmeldo:
It looks like I will not be able to get to this any time soon. Closing for the time being.

PR #188, opened 2021-03-19
Legends: Integration test

mmeldo:
Also improved definition of decks in tests.


server/game/effects.js
```diff
@@ -136,14 +136,15 @@ const Effects = {
             }
         };
     },
-    setSetupGold: function(value) {
```
tybar, line 139: Did this function get deleted intentionally? Or moved? I might just be missing it, but I don't see it in the rest of the file changes.

server/game/effects.js
```diff
@@ -136,14 +136,15 @@ const Effects = {
...
+    setMaxBullets: function(value) {
```
tybar, line 139: Should this function have a unit test associated with it?

tybar (changes requested):

server/game/effects.js
mmeldo: That is just a remainder from the throneteki code which used setup gold. I noticed it while working on this and removed.

server/game/effects.js
mmeldo: Yes, max bullets overall would probably deserve own unit test. Honestly, lot of effects are missing tests.

mmeldo: Created a new issue to handle tests for effects: #193

test/server/integration/legend.spec.js
```diff
@@ -0,0 +1,72 @@
+describe('Legend', function() {
```
tybar, line 1: I *think* this could use a couple more tests, though I'm definitely willing to discuss it:

1. There's nothing that checks that Doc Holliday correctly reduces your starting ghost rock. This one feels pretty important as it's a fairly unique modifier.
2. I think there should be a use of the Legend's ability.

mmeldo: 1. There is such test:
```
                it('should decrease wealth', function() {
                    expect(this.player1.player.ghostrock).toBe(18);
                });
```

2. Ability is really not that important, it is just a card with ability. I think what is more important is that if it has effect, it correctly affects cards in play. And there is a test for that.

mmeldo:
After some thinking, @tybar is right that we should add test to check legend's ability. Specifically shootout ability is available even though legend is not really in shootout.

PR #198, opened 2021-03-23
Implemented Point Blank

AaronSchnabel:
Took me way too long to get this working properly, but I've finally done it... now just time to wait and see what I have to change. :D


server/game/AbilityTarget.js
```diff
@@ -79,6 +79,10 @@ class AbilityTarget {
             return [context.player.getOpponent()];
         }
 
+        if(this.choosingPlayer === 'opponent') {
+            return [context.player.getOpponent()];
+        }
+
```
mmeldo, line 85: You did not need this at the end, but I believe it will be needed in the future so we keep it.

server/game/cards/01-DTR/PointBlank.js
```diff
@@ -0,0 +1,37 @@
...
+                    card.getType() === 'dude' &&
+                    card.isStud() && 
+                    card.isParticipating()
+                )
+            ],
+            handler: context => {
+                this.game.promptForSelect(context.player.getOpponent(), {
+                    activePromptTitle: 'Select a dude to ace',
+                    waitingPromptTitle: 'Waiting for opponent to select dude',
+                    cardCondition: { location: 'play area', controller: context.player.getOpponent(), participating: true, condition: card => card.bullets < context.costs.boot.bullets },
```
mmeldo, line 21: This could be done in 2 ways and you selected one of them so no problem here. Just FYI the second would be to pass `context` to properties of `promptForSelect` and then using `controller: 'opponent'` in cardCondition.
It would look like this:
```
// ...
waitingPromptTitle: 'Waiting for opponent to select dude',
context: context,
cardCondition: { location: 'play area', controller: 'opponent', participating: true, condition: card => card.bullets < context.costs.boot.bullets },
// ...
```

mmeldo (commented):
Will do some more testing and approve. You see, no changes needed 😃 Good job 👍

mmeldo (dismissed):

mmeldo (approved):

PR #201, opened 2021-03-24
Implemented The Winning Agenda

AaronSchnabel:
Also updated discardFromHand to allow prompt titles to be passed in.


server/game/cards/11-TCaR/TheWinningAgenda.js
```diff
@@ -0,0 +1,52 @@
...
+                cardCondition: { location: 'play area', condition: card => card.influence <= this.controller.hand.length },
+                cardType: 'dude'
+            },
+            handler: context => {
+                /* We have to cap the maximum number to 3 */
+                let discardAmount = context.target.influence;
+                if(discardAmount > 3) {
+                    discardAmount = 3;
+                }
+                context.player.discardFromHand(discardAmount, () => true, { title: this.title,
```
mmeldo, line 20: There is one small problem with this, but that is related to the `discardFromHand` in general. The discardAmount is **up to** amount, not exactly. Also, seeing in how many places Doomtown uses discard X cads to draw X cards, I guess we should make a `redraw` game action. If you do not mind, I will create that action using your PR.

mmeldo (changes requested):

server/game/cards/11-TCaR/TheWinningAgenda.js
AaronSchnabel: Ah, I didn't realize it was up to. Yeah, sounds good to make a new action. To me it makes sense to have a discard action that forces you to discard a specific amount of cards, and then with a parameter, redraw that same amount. That way you can do just the discard or discard and draw.

mmeldo (approved):

PR #204, opened 2021-03-24
Card recruitment drive

olafaxe:


mmeldo (changes requested):
Only really small things:
 - to be consistent we decided not to put dots at the end of messages
 - when you want to use player who is executing the action, use `context.player`. In most cases it will be the same as the `leader.owner`, but leader can theoretically be puppeted (controlled by other player)

mmeldo (approved):
Good job 👍 And have some more fun 😄

PR #210, opened 2021-03-26
Implemented Devil's Six Gun

AaronSchnabel:
Also changed cardselector.js so that Jokers can be selected by prompts.

### DIFF
--- a/server/game/cards/11-TCaR/DevilsSixGun.js
+++ b/server/game/cards/11-TCaR/DevilsSixGun.js
@@ -35,7 +35,7 @@ class DevilsSixGun extends GoodsCard {
                 );
                 this.game.promptForSelect(context.player, {
                     activePromptTitle: 'Select a card from your draw hand to discard',
-                    waitingPromptTitle: 'Waiting for ' + context.player + ' to discard a card from their draw hand',
+                    waitingPromptTitle: 'Waiting for opponent to discard a card from their draw hand',
                     cardCondition: { location: 'draw hand' },
                     onSelect: (player, cardToDiscard) => {
                         this.game.resolveGameAction(GameActions.discardCard({card: cardToDiscard}, context)).thenExecute(() => {
--- a/server/game/cards/11-TCaR/DevilsSixGun.js
+++ b/server/game/cards/11-TCaR/DevilsSixGun.js
@@ -1,0 +1,62 @@
+const GoodsCard = require('../../goodscard.js');
+const GameActions = require('../../GameActions/index.js');
+let selectedJoker = null;
+
+class DevilsSixGun extends GoodsCard {
+    setupCardAbilities(ability) {
+        this.whileAttached({
+            effect: ability.effects.setAsStud(this.uuid)
+        });
+        this.action({
+            title: 'Devil\'s Six Gun',
+            cost: ability.costs.bootSelf(),
+            playType: 'cheatin resolution',
+            handler: context => {
+                this.game.resolveGameAction(
+                    GameActions.search({
+                        title: 'Select a Joker',
+                        match: { condition: card => card.getType() === 'joker', location: ['draw deck', 'discard pile', 'dead pile'] },
+                        numToSelect: 1,
+                        message: {
+                            format: '{player} uses {source} to search for {searchTarget} and puts it into their draw hand'
+                        },
+                        cancelMessage: {
+                            format: '{player} uses {source} to search their draw deck, discard pile and Boot Hiil, but does not find a Joker'
+                        },
+                        handler: card => {
+                            context.player.moveCard(card, 'draw hand');
+                            selectedJoker = card;
+                        }
+                    }), {
+                        game: this.game,
+                        player: context.player,
+                        source: this
+                    }
+                );
+                this.game.promptForSelect(context.player, {
+                    activePromptTitle: 'Select a card from your draw hand to discard',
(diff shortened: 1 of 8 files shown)


server/game/cards/11-TCaR/DevilsSixGun.js
```diff
@@ -0,0 +1,62 @@
+const GoodsCard = require('../../goodscard.js');
+const GameActions = require('../../GameActions/index.js');
+let selectedJoker = null;
```
tybar, line 3: Should this variable be pushed down into the action? It isn't used outside of the action.

server/game/cards/11-TCaR/DevilsSixGun.js
```diff
@@ -0,0 +1,62 @@
...
+
+class DevilsSixGun extends GoodsCard {
```
tybar, line 5: Probably a more long-term thing, but should we have a GadgetCard extend GoodsCard?

Asking now as that decision will probably be good to make earlier rather than later.

server/game/cards/11-TCaR/DevilsSixGun.js
mmeldo: Yes, if you want to use variable in cards across functions, it is better to use `this.variable`. However, I have reworked the discard event and updated this PR so the variable is no longer needed.

server/game/cards/11-TCaR/DevilsSixGun.js
mmeldo: We do not need `GadgetCard` as gadget is easily identified by the keyword and does not have any specific functions that should be put into the new class.

server/game/cards/11-TCaR/DevilsSixGun.js
```diff
@@ -0,0 +1,66 @@
+const GoodsCard = require('../../goodscard.js');
+const GameActions = require('../../GameActions/index.js');
+
+class DevilsSixGun extends GoodsCard {
+    setupCardAbilities(ability) {
+        this.whileAttached({
+            effect: [
+                ability.effects.setAsStud(this.uuid),
+                ability.effects.modifyBullets(this.bullets)
```
mmeldo, line 9: I have added bullets effects as Devil Six Gun has +1 bullets.

server/game/cards/11-TCaR/DevilsSixGun.js
```diff
@@ -0,0 +1,66 @@
...
+                        numToSelect: 1,
+                        message: {
+                            format: '{player} uses {source} to search for {searchTarget} and puts it into their draw hand'
+                        },
+                        cancelMessage: {
+                            format: '{player} uses {source} to search their draw deck, discard pile and Boot Hill, but does not find a Joker'
+                        },
+                        handler: jokerCard => {
+                            context.player.moveCard(jokerCard, 'draw hand');
+                            this.game.promptForSelect(context.player, {
```
mmeldo, line 30: I have moved the discard select and event handling to the search handler. Thanks to this, we do not need global `selectedJoker` variable.

server/game/cards/11-TCaR/DevilsSixGun.js
```diff
@@ -0,0 +1,66 @@
...
+                                }                   
+                            });
+                            let eventHandler = event => {
+                                if(event.player.drawHand.find(card => jokerCard === card)) {
+                                    event.player.moveCard(jokerCard, 'draw deck');
+                                    event.player.shuffleDrawDeck();
+                                    this.game.addMessage('{0} shuffles Joker back to the draw deck thanks to {1}', event.player, this);
+                                }
+                            };
+                            this.game.once('onBeforeDrawHandDiscarded', eventHandler);
```
mmeldo, line 51: I realized that the `onJokerAced` event is: 
1. not really needed as we have `onCardAced` event
2. in this instance, joker is not really aced, but shuffled to draw deck instead so I created new event `onBeforeDrawHandDiscarded` that can be used to shuffle joker to draw deck before it is aced.

mmeldo (approved):
I have updated the event handling and when doing that, also fixed some other things. Please take a look, and test.

mmeldo:
Good job on that joker issue 👍

server/game/cards/11-TCaR/DevilsSixGun.js
AaronSchnabel: Thanks, I overlooked that.

server/game/cards/11-TCaR/DevilsSixGun.js
tybar: As someone not familiar with this pull request tool...did you doing that erase my comment?

mmeldo: yes, github automatically resolves review comments when the stuff you commented on is changed. it is not erased, just hidden/resolved.

mmeldo:
I think I should take a look at the event resolution ... it looks like the `then` handling is not happening after the event is emitted, but before. I would like to use this PR for that change as the Devil Six Gun is good card to test.

mmeldo:
Finally, I was able to make it work by introducing "beforeEventHandlers".  Anyone now can register beforeEventHandler by calling `game.before(eventName, beforeHandler, useOnce, condition)`
I believe there are other usages where cards refer to use before some event.
Can you please test if it works as intended?

mmeldo (dismissed):

mmeldo (approved):

PR #211, opened 2021-03-28
Hex Slingin

janosikm:
- Card implementation 
- supporting changes for discard from play cost


mmeldo (commented):
Since you are already in that part of code, could you also add costs to discard multiple from play?

mmeldo (approved):

PR #357, opened 2021-03-31
Card: Magical Distraction, Remy LaPointe, Mechanical Horse

mmeldo:
Small fixes for paying of reduceable ghost rock and move options.

### DIFF
--- a/server/game/cards/01-DTR/MagicalDistraction.js
+++ b/server/game/cards/01-DTR/MagicalDistraction.js
@@ -1,0 +1,29 @@
+const ActionCard = require('../../actioncard.js');
+
+class MagicalDistraction extends ActionCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Magical Distraction',
+            playType: ['cheatin resolution'],
+            cost: ability.costs.discardFromPlay(card => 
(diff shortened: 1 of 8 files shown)


server/game/cards/01-DTR/MagicalDistraction.js
```diff
@@ -0,0 +1,29 @@
...
+            cost: ability.costs.discardFromPlay(card => 
+                card.location === 'play area' &&
+                card.getType() === 'spell' &&
+                card.parent.controller === this.controller &&
+                card.parent.isParticipating()
+            ),
+            handler: context => {
+                context.player.pull((pulledCard, pulledValue) => {
+                    let currentRank = context.player.getTotalRank();
+                    context.player.modifyRank(pulledValue - currentRank);
```
janosikm, line 17: Should not we limit the hand rand to max of 11?

mmeldo: It is similar as with card stats. We keep the overall hand rank which can go over 11, but when used, we use max of 11. So when you do `getTotalRank` for a player, there is a condition:
```
if(totalRank > 11) {
    return 11;
}
```

PR #360, opened 2021-04-02
Implemented Phantasm

AaronSchnabel:


server/game/cards/03-FF/Phantasm.js
```diff
@@ -0,0 +1,93 @@
+const GameActions = require('../../GameActions/index.js');
+const SpellCard = require('../../spellcard.js');
+
+class Phantasm extends SpellCard {
+    setupCardAbilities(ability) {
+        this.spellAction({
+            title: 'Noon: Phantasm: Move Unbooted Dude',
```
mmeldo, line 7: I do not think we need name of the card in title as it is displayed in the prompt anyway.

server/game/cards/03-FF/Phantasm.js
```diff
@@ -0,0 +1,93 @@
...
+                        this.game.addMessage('{0} uses {1} to move {2} to {3}', context.player, this, context.target, location);
+                        return true;
+                    }
+                });
+            },
+            source: this
+        });
+
+        this.spellAction({
+            title: 'Noon: Phantasm: Move Booted Dude',
```
mmeldo, line 49: I do not think we need name of the card in title as it is displayed in the prompt anyway.

server/game/cards/03-FF/Phantasm.js
```diff
@@ -0,0 +1,93 @@
...
+            cost: ability.costs.bootSelf(),
+            target: {
+                activePromptTitle: 'Select unbooted dude to move',
+                cardCondition: { 
+                    location: 'play area', 
+                    controller: 'opponent', 
+                    condition: card => 
+                        card.gamelocation === this.gamelocation ||
+                        card.getGameLocation().isAdjacent(this.gamelocation) &&
+                        !card.booted
```
mmeldo, line 18: This condition lets you choose also booted dudes at the caster's location that is not correct. That is because && has precedence over ||.
The condition should be:

```suggestion
                        (card.gamelocation === this.gamelocation ||
                        card.getGameLocation().isAdjacent(this.gamelocation)) &&
                        !card.booted
```

mmeldo (changes requested):

mmeldo (approved):

PR #361, opened 2021-04-04
Adding Test Plan draft

tybar:
First pull request!

I am not sure what all guidance people will need, so I can make this more granular as needed. I wanted to start out high level first, and drill down as needed.

Also, please let me know if there's some fancier Markdown that would look better. :)

### DIFF
--- a/docs/test-plan
+++ b/docs/test-plan
@@ -1,0 +1,26 @@
+#Test plan
+
+### Philosophy
+* It's cheaper to bake in quality early than trying to sprinkle it on top later.
+* Test at the lowest level that is responsible. Write unit tests over integration tests, integration tests over system level tests.
+* Automate what you can. We will probably end up with some manual test scripts, but they should be relatively small in number.
+* Flaky tests just make people distrust the tests. If you need help, get help to fix it.
+
+### What to test
+* Basic interactions of the game should be tested (callouts, playing action cards, etc.).
+* If a bug is reported, there should be a test checked in as part of the bug fix.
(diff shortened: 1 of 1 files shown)


docs/test-plan
```diff
@@ -0,0 +1,26 @@
...
+* If in doubt, write a test. You'll never get yelled at for having one too many tests in your code. :)
+
+### What not to test
+* Not every card needs a test. This includes plain dudes without any card text, as well as relatively simple cards like Tusk.
+
+### How to test
+* Start by identifying the main "happy" path for the code you're writing.
+* Then identify any edge cases that are relevant for the you're writing.
+* Finally, start building up tests. Your tests should test functionality, not implementation (i.e. they should not need to be updated because of internal changes to what code you're writing)
+
```
tybar, line 22: Should have put in something to the effect of "make sure to ad-hoc new cards and interactions with the client too."

mmeldo:
Congrats to your first pull request. The plan covers almost all, but I believe in current situation with only me doing engine changes, we do not need such broad plan. What I would like to do now is identify areas that are not properly covered by tests. There are already some issues with `tests` label to identify such areas. And of course to implement those tests afterwards.

mmeldo (commented):
Give the `test-plan` file `.md` extension so GitHub knows it should treat it as markdown.

tybar:
> Give the `test-plan` file `.md` extension so GitHub knows it should treat it as markdown.

Done.

PR #363, opened 2021-04-06
Cannot move and boot effects, refactor and some cards

mmeldo:
Cards: Yagn's Mechanical Skeleton, Scoop Hound, Angelica Espinosa

I am aware this one will need some unit/integration tests, but I would like to do it as separate PR as this one is already quite big.

### DIFF
--- a/docs/cardpool-status.md
+++ b/docs/cardpool-status.md
@@ -1,6 +1,6 @@
 # Card implementation status
 
-![Cards implemented](https://progress-bar.dev/141/?scale=875&title=implemented%20&width=200&suffix=%20cards) out of 875
+![Cards implemented](https://progress-bar.dev/144/?scale=875&title=implemented%20&width=200&suffix=%20cards) out of 875
 
 **Legend**
 
@@ -195,7 +195,7 @@
 |                     | 03004| Gang Yi
 |                     | 03005| Tallulah "Lula" Morgan
 |                     | 03006| Dr. Emanuel Ashbel
-|                     | 03007| Angélica Espinosa
+|:heavy_check_mark:   | 03007| Angélica Espinosa
 |                     | 03008| José Morales
 |                     | 03009| Genesee "Gina" Tailfeathers
 |                     | 03010| Mario Crane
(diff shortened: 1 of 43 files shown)


AaronSchnabel:
I got the attached error when attempting to use Phantasm to move an unbooted dude from their home to TS
[movement error Phantasm.txt](https://github.com/townteki/townsquare/files/6272900/movement.error.Phantasm.txt)

AaronSchnabel:
I tried to use Pistol Whip on a dude who had Yagn's attached, and it properly prevented the effect from taking place, but I was able to target the dude, and the game message said "drcthulhu2 uses Pistol Whip to send home booted " I don't know if there is a way to add a message on the cannot properties when it prevents something from happening, or if we should just prevent somebody from being targeted by an ability if they have a cannot effect for that ability.

AaronSchnabel:
I attached Yagn's to Jarret Blake, and when I tried to use his shootout ability to move him into the posse, I got the needToBoot error again.
[need to boot error.txt](https://github.com/townteki/townsquare/files/6273168/need.to.boot.error.txt)

AaronSchnabel:
Yagn's Mechanical Skeleton is broken somehow. If you attach it to a dude, everything works fine for that day. However, when you move into a new day, it makes you pay all of your money as upkeep (setting your stash to 0), and then it breaks something so that you cannot use the buttons to add ghost rock to your stash. This means you can't do any shoppin, etc. Using the /remove-from-game command to remove Yagn's and then moving to a new day does not reset this.

mmeldo:
> Yagn's Mechanical Skeleton is broken somehow. If you attach it to a dude, everything works fine for that day. However, when you move into a new day, it makes you pay all of your money as upkeep (setting your stash to 0), and then it breaks something so that you cannot use the buttons to add ghost rock to your stash. This means you can't do any shoppin, etc. Using the /remove-from-game command to remove Yagn's and then moving to a new day does not reset this.

This was not problem with Yagn per se, but with the way upkeep has been determine. Yagn's bad luck was that it was first card other than dudes to have influence. Should be fixed.

mmeldo:
All other reported bugs are fixed plus one new.

mmeldo:
Had a short testing game with @janosikm and everything related to this PR went without problems.
@AaronSchnabel please verify the bug fixes for the problems you reported.

AaronSchnabel:
Everything seems to be working A-OK to me. Good job!

PR #365, opened 2021-04-06
Implemented Consecration

AaronSchnabel:
Pulled in changes from townteki/townsquare/Update-casualties-tracking


server/game/gamesteps/shootout.js
```diff
@@ -34,6 +34,7 @@ class Shootout extends Phase {
             this.opposingPosse = new ShootoutPosse(this, this.opposingPlayer, false);
         }
 
+        this.winnerCasualtiesMod = 0;
```
mmeldo, line 37: This changes in `shootout.js` are no longer needed.

mmeldo (changes requested):

server/game/gamesteps/shootout.js
AaronSchnabel: Bah, thought I got all that cleaned up.

mmeldo: just a reminder that this is waiting only for the cleanup. sorry if you have that already covered :)

server/game/cards/04.2-NTB/Consecration.js
```diff
@@ -0,0 +1,39 @@
...
+            title: 'Consecration',
+            playType: 'cheatin resolution',
+            target: {
+                activePromptTitle: 'Select your dude to Consecrate',
+                waitingPromptTitle: 'Waiting for opponent to select a dude',
+                cardCondition: { location: 'play area', controller: 'current'},
+                cardType: ['dude']
+            },
+            difficulty: 7,
+            actionContext: { card: this, gameAction: ['increaseBullets', 'increaseInfluence', 'setAsStud'] },
```
mmeldo, line 15: This will cause that if the parent cannot be affected even by one of the effects, all will be prevented. Is that ok or should be only the one effect prevented?
For example if blessed cannot have bullets increased, not only the bullet increase will be prevented, but also influence and stud effect.

server/game/cards/04.2-NTB/Consecration.js
```diff
@@ -0,0 +1,39 @@
...
+            title: 'Consecration',
+            playType: 'cheatin resolution',
+            target: {
+                activePromptTitle: 'Select your dude to Consecrate',
+                waitingPromptTitle: 'Waiting for opponent to select a dude',
+                cardCondition: { location: 'play area', controller: 'current'},
+                cardType: ['dude']
+            },
+            difficulty: 7,
+            actionContext: { card: this, gameAction: 'increaseBullets' || 'increaseInfluence' || 'setAsStud' },
```
mmeldo, line 15: The `gameAction` does not work that way (with "or" ||). I saw the discussion on discord, so you can remove the `actionContext` here completely. If some effect is prevented, it will be so no need for `actionContext`

mmeldo (changes requested):

mmeldo (approved):

PR #367, opened 2021-04-06
Implemented Confession

AaronSchnabel:


server/game/cards/04.1-FJ/Confession.js
```diff
@@ -0,0 +1,65 @@
...
+                this.game.resolveGameAction(GameActions.bootCard({ card: this.parent }), context);
+                this.game.resolveGameAction(GameActions.addBounty({
+                    card: context.target, amount: this.parent.getSkillRating('blessed')
+                }));
+                this.game.addMessage('{0} uses {1} to boot {2} and add bounty to {3}', context.player, this, this.parent, context.target);
+            },
+            source: this
+        });
+
+        this.spellAction({
```
mmeldo, line 33: Looks like this spell was just copied from the other action. It cannot work because it has a function typo (ModifyGhostRock) and gameAction is incorrect.

server/game/cards/04.1-FJ/Confession.js
```diff
@@ -0,0 +1,65 @@
...
+                    location: 'play area', 
+                    condition: card =>
+                        card.gamelocation === this.gamelocation || 
+                        card.getGameLocation().isAdjacent(this.gamelocation)
+                },
+                cardType: ['dude'],
+                gameAction: 'addBounty'
+            },
+            onSuccess: (context) => {
+                this.game.resolveGameAction(GameActions.bootCard({ card: this.parent }), context);
```
mmeldo, line 24: I know it is not strictly a cost, but since it has to be fulfilled, I would put the booting of parent to costs.

server/game/cards/04.1-FJ/Confession.js
```diff
@@ -0,0 +1,65 @@
+const GameActions = require('../../GameActions/index.js');
+const SpellCard = require('../../spellcard.js');
+
+class Confession extends SpellCard {
+    setupCardAbilities(ability) {
+        this.spellAction({
+            title: 'Noon: Miracle 6',
```
mmeldo, line 7:
```suggestion
            title: 'Noon: Add bounty to a dude',
```
I would use more descriptive action titles.

server/game/cards/04.1-FJ/Confession.js
```diff
@@ -0,0 +1,65 @@
...
+                this.game.resolveGameAction(GameActions.addBounty({
+                    card: context.target, amount: this.parent.getSkillRating('blessed')
+                }));
+                this.game.addMessage('{0} uses {1} to boot {2} and add bounty to {3}', context.player, this, this.parent, context.target);
+            },
+            source: this
+        });
+
+        this.spellAction({
+            title: 'Noon: Miracle 7',
```
mmeldo, line 34:
```suggestion
            title: 'Noon: Transfer bounty from a dude',
```
I would use more descriptive action titles.

mmeldo (changes requested):

server/game/cards/04.1-FJ/Confession.js
AaronSchnabel: I didn't put it as a cost, since it only happens if the spell pull is successful. Unless I misunderstand, if I put it as a cost, it will boot the parent even if the spell pull is unsuccessful. I can add something to the condition to make sure that the parent is not already booted though, to make sure that it can happen.

mmeldo: You are right .. but there can be a cannot effect on the parent, so at least I would put the message in `thenExecute`

server/game/cards/04.1-FJ/Confession.js
```diff
@@ -0,0 +1,65 @@
...
+                },
+                cardType: ['dude'],
+                gameAction: 'removeBounty'
+            },
+            onSuccess: (context) => {
+                this.game.resolveGameAction(GameActions.removeBounty({
+                    card: context.target, amount: 1
+                }));
+                context.player.modifyGhostRock(1);
+                this.game.addMessage('{0} uses {1} to take 1 bounty from {2} and add it to their stash', context.player, this, context.target);
```
mmeldo, line 56: I would put the modifyGhostRock and message in `thenExecute` after the removeBounty game action. In case anything prevents the removeBounty, we do not want the ghost rock to be added.

mmeldo (changes requested):

server/game/cards/04.1-FJ/Confession.js
AaronSchnabel: Good point, will do.

server/game/cards/04.1-FJ/Confession.js
```diff
@@ -0,0 +1,70 @@
...
+                cardType: ['dude'],
+                gameAction: 'removeBounty'
+            },
+            onSuccess: (context) => {
+                this.game.resolveGameAction(GameActions.removeBounty({
+                    card: context.target, amount: 1
+                })).thenExecute(() => {
+                    context.player.modifyGhostRock(1).thenExecute(() => {
+                        this.game.addMessage('{0} uses {1} to take 1 bounty from {2} and add it to their stash', context.player, this, context.target);
+                    });
```
mmeldo, line 60:
```suggestion
                    context.player.modifyGhostRock(1);
                    this.game.addMessage('{0} uses {1} to take 1 bounty from {2} and add it to their stash', context.player, this, context.target);
```
Function `modifyGhostRock` cannot have `thenExecute` as it does not return Event object.

mmeldo (changes requested):

mmeldo (approved):

PR #429, opened 2021-05-04
increment zeromq version

stopfstedt:
i was getting build errors when this dependency was compiling from source during the build process (the same holds true when spinning up the included docker build).

bumping zeromq seems to overcome this issue.

for reference, this is my environment: 

```
$ npm --version
6.14.12

$ node --version
v14.16.1

$ cat /etc/issue
Ubuntu 20.04.2 LTS \n \l
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1461,44 +1461,6 @@
         "file-uri-to-path": "1.0.0"
       }
     },
-    "bl": {
-      "version": "1.2.2",
-      "resolved": "http://registry.npmjs.org/bl/-/bl-1.2.2.tgz",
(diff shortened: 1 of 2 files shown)


mmeldo:
There are developers working on Windows which seems to be able to run only on Node 10.x, and also someone was setting up on Mac. Because of this, I am reluctant to change any dependencies in the github repo now. Later, there is a plan to update all the dependencies because there are also some security warnings by dependabot. 
Rather than changing the dependencies in the main repository, I would update the the contributing guide and open an issue as a reminder we should bump the dependencies at some point.

mmeldo:
And welcome 😄 Looking forward to your contributions. In case you want to start working on some card or feature, just put a comment there and someone will assign it to you.

PR #440, opened 2021-05-07
Sequoia

AaronSchnabel:


server/game/cards/12-2T2D/AlexanderSequoia.js
```diff
@@ -0,0 +1,30 @@
...
+
+class AlexanderSequoia extends DudeCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            condition: () => this.isInTownSquare(),
+            effect: ability.effects.modifySundownDiscard(1)
+        });
+        
+        this.action({
+            title: 'Noon',
```
mmeldo, line 12: Use at least the card name, or some short description as Title, because it is possible some other action can be added to a dude (e.g. via Nicodemus Whateley), and then the title is used in the menu and "Noon" is not very good.

server/game/cards/12-2T2D/AlexanderSequoia.js
```diff
@@ -0,0 +1,30 @@
...
+            playType: 'noon',
+            ifCondition: () => this.isInTownSquare(),
+            ifFailMessage: context => {
+                this.game.addMessage('{0} uses {1}\'s ability but {1} does not unboot because he is not in town square', context.player, this);
+            },
+            cardCondition: this.booted,
```
mmeldo, line 18: Add also `actionContext: { card: this, gameAction: 'unboot' }`. Just in case there will be some effect introduced that prevents unbooting.
```suggestion
            cardCondition: this.booted,
            actionContext: { card: this, gameAction: 'unboot' },
```

mmeldo (changes requested):

mmeldo (approved):

PR #443, opened 2021-05-08
108 Gracious Gifts

AaronSchnabel:
Need help with the persistent effect. Tried removing the ghost rock check just to see if it would work at sundown and it still didn't allow discarding more than one card.

### DIFF
--- a/server/game/cards/12-2T2D/108GraciousGifts.js
+++ b/server/game/cards/12-2T2D/108GraciousGifts.js
@@ -1,0 +1,38 @@
+const GameActions = require('../../GameActions/index.js');
+const OutfitCard = require('../../outfitcard.js');
+
+class GraciousGifts extends OutfitCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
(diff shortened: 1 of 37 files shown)


server/game/cards/12-2T2D/108GraciousGifts.js
```diff
@@ -0,0 +1,37 @@
+const GameActions = require('../../GameActions/index.js');
+const OutfitCard = require('../../outfitcard.js');
+
+class GraciousGifts extends OutfitCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            condition: () => this.game.currentPhase === 'sundown' && (this.owner.getSpendableGhostRock() < this.owner.opponent.getSpendableGhostRock()),
```
mmeldo, line 7: There is no `opponent` attribute for player, you need to use `getOpponent()` function. Also, the `modifySundownDiscard` effect targets player, so you need to put a `match: this.owner`.
Last thing is that I am not sure about the `getSpendableGhostRock` as it returns GR also from other potential GR sources. I am not sure if the 108 Gracious gifts refers only to player stash, or his overall GR pool.

mmeldo:
Pushed a fix for persistent effects of outfits. Your is a first outfit card with a persistent effect :)

mmeldo (approved):

PR #444, opened 2021-05-08
Ghostlycommunion

AaronSchnabel:
Feel like I'm in the weeds on this one, not sure I'm even on the right track on how to solve this one. Could use some direction here.


server/game/cards/09.3-TSs/GhostlyCommunion.js
```diff
@@ -0,0 +1,45 @@
...
+    setupCardAbilities(ability) {
+        this.spellAction({
+            title: 'Noon: Move Dude',
+            playType: 'noon',
+            cost: ability.costs.bootSelf(),
+            difficulty: 5,
+            target: {
+                activePromptTitle: 'Choose a location to move your dude to',
+                waitingPromptTitle: 'Waiting for opponent to choose a location',
+                cardCondition: card => {card.location === 'play area' && (card.hasKeyword('holy ground') || card.adjacentLocations().hasKeyword('holy ground'))},
```
mmeldo, line 14: You can have a `cardCondition` as a function or object, but you seems to be mixing both.
For target, I usually use object, you can check some other cards:
``` 
cardCondition: { location: 'play area', condition: card => card.hasKeyword('holy ground') || card.adjacentLocations().some(location => location.locationCard.hasKeyword('holy ground') }
```
You see the other problem .. you were treating the `card.adjacentLocations()` result as a card, but it is array.

server/game/cards/09.3-TSs/GhostlyCommunion.js
```diff
@@ -0,0 +1,45 @@
...
+            source: this
+        });
+
+        this.spellAction({
+            title: 'Shootout: Join Posse',
+            playType: 'shootout:join',
+            cost: ability.costs.bootSelf(),
+            difficulty: 7,
+            condition: () => this.game.shootout &&
+                (this.game.shootout.shootoutLocation.hasKeyword('holy ground') || this.game.shootout.shootoutLocation.adjacentLocations().hasKeyword('holy ground')),
```
mmeldo, line 32: Same as with the other action, you are treating the `card.adjacentLocations()` result as a card, but it is array.

server/game/cards/09.3-TSs/GhostlyCommunion.js
```diff
@@ -0,0 +1,45 @@
...
+            onSuccess: (context) => {
+                this.game.resolveGameAction(GameActions.joinPosse({ card: this.parent }), context);
+                this.parent.setAsStud();
```
mmeldo, line 35: Giving of a stud is effect, you cannot do it like this. If the card text does not have special timing effect, you should use the `applyAbilityEffect` function. See for example card `TheStakesJustRose`.

mmeldo (approved):

PR #445, opened 2021-05-09
implements B&B Attorneys.

### LINKED ISSUE #165: DTR - B & B Attorneys

stopfstedt:
fixes #165

### DIFF
--- a/server/game/cards/01-DTR/BBAttorneys.js
+++ b/server/game/cards/01-DTR/BBAttorneys.js
@@ -8,7 +8,7 @@ class BBAttorneys extends DeedCard {
             playType: ['noon'],
             cost: ability.costs.bootSelf(),
             target: {
-                activePromptTitle: 'Select a wanted dude to raise or lower their bounty by one.',
+                activePromptTitle: 'Select a wanted dude',
                 cardCondition: { location: 'play area', wanted: true },
                 cardType: ['dude']
             },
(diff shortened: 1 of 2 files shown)


server/game/cards/01-DTR/BBAttorneys.js
```diff
@@ -0,0 +1,62 @@
...
+const GameActions = require('../../GameActions/index.js');
+
+class BBAttorneys extends DeedCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'B & B Attorneys',
+            playType: ['noon'],
+            cost: ability.costs.bootSelf(),
+            target: {
+                activePromptTitle: 'Select a wanted dude to raise or lower their bounty by one.',
```
mmeldo, line 11:
```suggestion
                activePromptTitle: 'Select a wanted dude',
```
I would make it shorter, you should know what you are doing based on the cards you are using.

mmeldo (changes requested):
Great first card 👍 Even though I see the "good first card" was not really correct in this case. I would suggest just a small change in the target select title.

server/game/cards/01-DTR/BBAttorneys.js
stopfstedt: handled. :heavy_check_mark:

mmeldo (approved):

PR #447, opened 2021-05-09
adds discard-to-hand-size step to sundown phase.

### LINKED ISSUE #446: increased hand size never resets during sundown if persistent effect that was causing it expires
**Describe the bug**

**To Reproduce**
1. Start a game with Randall in your starting posse
2. Observe that hand size is increased to six
3. Move Randall to town square by himself and leave him there by himself
4. Pass through to Sundown
5. Do _not_ discard a card to draw a card
6. Start the new day with six cards in hand. 

This is incorrect. The rulebook states in section 5.4 Sundown:

> All players may choose to discard one card from their play hand, with the Winner choosing first. All players then discard down or refill their play hand to their maximum hand size (normally 5 cards).  

**Screenshots**

handsize6

**Additional context**
I encountered this issue when working on implementing #436 and then verified it being a bug in the game engine (and not the card implementation itself) by creating a similar incorrect game state using Randall.

stopfstedt:
fixes #446 

based on the [discard to meet reserve prompt](https://github.com/throneteki/throneteki/blob/master/server/game/gamesteps/taxation/DiscardToReservePrompt.js) from throneteki.

### DIFF
--- a/server/game/gamesteps/sundown/discardtohandsizeprompt.js
+++ b/server/game/gamesteps/sundown/discardtohandsizeprompt.js
@@ -1,0 +1,53 @@
+const BaseStep = require('../basestep');
+
+class DiscardToHandSizePrompt extends BaseStep {
+    constructor(game) {
+        super(game);
+        this.players = this.game.getPlayersInFirstPlayerOrder();
+    }
+
+    continue() {
+        while(this.players.length) {
+            let currentPlayer = this.players.shift();
+            if(! currentPlayer.isOverHandsizeLimit()) {
+                this.game.addMessage('{0} already meets their hand size limit', currentPlayer);
+            } else {
+                this.promptPlayerToDiscard(currentPlayer);
+                return false;
+            }
+        }
+        this.game.raiseEvent('onHandsizeLimitChecked');
+    }
+
+    promptPlayerToDiscard(currentPlayer) {
+        let overHandsize = currentPlayer.hand.length - currentPlayer.handSize;
+        let cardOrCards = overHandsize === 1 ? 'card' : 'cards';
+        this.game.promptForSelect(currentPlayer, {
+            ordered: true,
+            mode: 'exactly',
+            numCards: overHandsize,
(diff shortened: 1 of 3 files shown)


server/game/gamesteps/sundown/discardtohandsizeprompt.js
```diff
@@ -0,0 +1,53 @@
...
+            }
+        }
+        this.game.raiseEvent('onHandsizeLimitChecked');
+    }
+
+    promptPlayerToDiscard(currentPlayer) {
+        let overHandsize = currentPlayer.hand.length - currentPlayer.handSize;
+        let cardOrCards = overHandsize === 1 ? 'card' : 'cards';
+        this.game.promptForSelect(currentPlayer, {
+            ordered: true,
```
mmeldo, line 26: I do not think there needs to be ordered discard at this time. I believe the only instance where it makes sense is for LeMat.

server/game/gamesteps/sundown/discardtohandsizeprompt.js
```diff
@@ -0,0 +1,53 @@
...
+    constructor(game) {
+        super(game);
+        this.players = this.game.getPlayersInFirstPlayerOrder();
+    }
+
+    continue() {
+        while(this.players.length) {
+            let currentPlayer = this.players.shift();
+            if(! currentPlayer.isOverHandsizeLimit()) {
+                this.game.addMessage('{0} already meets their hand size limit', currentPlayer);
```
mmeldo, line 13: I am not sure if we should add message even if the player meets the hand size limit. The discard down is not very often so I think only the message when it happens is needed.

mmeldo (commented):
Just a few comments, open for discussion.

server/game/gamesteps/sundown/discardtohandsizeprompt.js
stopfstedt: :heavy_check_mark:

server/game/gamesteps/sundown/discardtohandsizeprompt.js
stopfstedt: :heavy_check_mark:

PR #451, opened 2021-05-10
Card: Emre, The Turkish Bear

mmeldo:


server/game/player.js
```diff
@@ -1220,23 +1220,24 @@ class Player extends Spectator {
...
-                this.game.queueStep(new JokerPrompt(this.game, pulledCard, callback));
-            } else {
-                callback(pulledCard, pulledCard.value, pulledCard.suit);
+        this.game.raiseEvent('onCardPulled', { card: pulledCard, props }, event => {
+            if(callback) {
+                if(event.card.getType() === 'joker') {
+                    this.game.queueStep(new JokerPrompt(this.game, event.card, callback));
+                } else {
+                    callback(event.card, event.card.value, event.card.suit);
+                }
```
mmeldo, line 1238: I put the whole logic that should be executed after pull (callback) inside the event handler so it can be replaced by the Emre's ability. It will keep the pull event handler the same, but will feed a different pulledCard into the callback.

server/game/gamesteps/abilityresolver.js
```diff
@@ -231,7 +231,7 @@ class AbilityResolver extends BaseStep {
             this.game.resolveEvent(event);
         }
         if(this.context.pull) {
-            this.context.player.handlePulledCard(this.context.pull.pulledCard);
+            this.context.pull.pulledCard.owner.handlePulledCard(this.context.pull.pulledCard);
```
mmeldo, line 234: Changed the player doing the pulled card handling to the owner of the card so it will go to proper discard pile. Same was done in the `player.handlePull` function.

stopfstedt (approved):
LGTM. for testing purposes, i replaced `technique` with `hex` (saw that in the last commit) and it worked like a charm. the joker pull edge-case is also correctly handled.

stopfstedt:
ok, so might have hit a snag after all:

when Emre pulls a Joker off the top, it goes into this "played/pulled" area on the board. it seems to be stuck there, can't really interact with it. shouldn't the Joker pull have triggered a prompt instead?

image

stopfstedt (changes requested):
pulling replacement for a Joker via Emre's effect does appear to resolve correctly. Please see my last comment in the conversation tab.

mmeldo:
Yeah, the joker situation is the one I forgot to test and it turned out to be little tricky. Thanks for catching it out.
I have pushed a fixed for it, tested and it seems the joker pulled by Emre situation is resolved correctly now.
During the testing I have found another problem with React abilities #467

stopfstedt (approved):
re-reviewed. LGTM now.

PR #454, opened 2021-05-10
Implemented Asakichi Cooke

AaronSchnabel:


server/game/cards/06-TLS/AsakichiCooke.js
```diff
@@ -0,0 +1,40 @@
...
+                activePromptTitle: 'Choose one of your other dudes at this location to move',
+                waitingPromptTitle: 'Waiting for opponent to choose a dude',
+                cardCondition: { 
+                    location: 'play area', 
+                    controller: 'current', 
+                    condition: card => card.locationCard === this.locationCard && card !== this,
+                    cardType: 'dude'
+                }
+            },
+            actionContext: { card: this, gameAction: 'moveDude' },
```
mmeldo, line 19: We are not moving Asakichi but a targeted dude, so the actionContext should not be used here. It is usually used only if the action is on the card itself or on the parent.
In this case, there should be `gameAction: 'moveDude'` in the target definition.

server/game/cards/06-TLS/AsakichiCooke.js
```diff
@@ -0,0 +1,40 @@
+const DudeCard = require('../../dudecard.js');
+const GameActions = require('../../GameActions/index.js');
+
+class AsakichiCooke extends DudeCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Move Your Dude',
+            cost: ability.costs.discardFromHand(),
+            target: {
+                activePromptTitle: 'Choose one of your other dudes at this location to move',
```
mmeldo, line 10: Let's use less descriptive titles, you should know what you are doing when triggering a card.
e.g. "Choose your dude at this location"

server/game/cards/06-TLS/AsakichiCooke.js
```diff
@@ -0,0 +1,40 @@
...
+            handler: context => {
+                this.game.promptForLocation(context.player, {
+                    activePromptTitle: 'Select where to move your dude',
+                    waitingPromptTitle: 'Waiting for opponent to select location',
+                    onSelect: (player, location) => {
+                        this.game.resolveGameAction(GameActions.moveDude({ 
+                            card: context.target, 
+                            targetUuid: location.uuid
+                        }), context);   
+                        this.game.addMessage('{0} uses {1} to discard a card from hand and move {2} to {3}.', player, this, context.target, location);
```
mmeldo, line 29:
```suggestion
                        this.game.addMessage('{0} uses {1} to discard a card from hand and move {2} to {3}', player, this, context.target, location);                                 
```
Removed the dot at the end.

mmeldo (changes requested):

PR #463, opened 2021-05-11
Implemented Bairds Build & Loan

### LINKED ISSUE #299: DD - Baird's Build and Loan

AaronSchnabel:
fixes #299

### DIFF
--- a/server/game/cards/02.2-DD/BairdsBuildLoan.js
+++ b/server/game/cards/02.2-DD/BairdsBuildLoan.js
@@ -1,0 +1,28 @@
+const DeedCard = require('../../deedcard.js');
+const StandardActions = require('../../PlayActions/StandardActions.js');
+
+class BairdsBuildLoan extends DeedCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Play a Deed',
+            playType: ['noon'],
+            cost: ability.costs.bootSelf(),
+
+            target:{
+                activePromptTitle: 'Select a deed',
+                cardCondition: card => card.location === 'hand' && (this.controller.ghostrock >= card.cost - 2),
+                cardType: ['deed']
+            },
+            message: context => {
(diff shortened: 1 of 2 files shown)


server/game/cards/02.2-DD/BairdsBuildLoan.js
```diff
@@ -0,0 +1,28 @@
...
+class BairdsBuildLoan extends DeedCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Play a Deed',
+            playType: ['noon'],
+            cost: ability.costs.bootSelf(),
+
+            target:{
+                activePromptTitle: 'Select a deed',
+                cardCondition: card => card.location === 'hand' && (this.controller.ghostrock >= card.cost - 2),
```
mmeldo, line 13: I would not use the ghostrock condition on this one. I know it can lead to situation when player chooses deed he cannot afford, but in that case nothing happens (Baird build&loan will be spent but that can be reset easily). However in the opposite situation it will have to played around.
By opposite situation I mean:
 - condition does not allow the deed to be put in play because of ghost rock
 - but there will be other reducers or ghost rock sources on other cards
 - these reducers or ghost rock sources would allow player to put the deed into play

mmeldo (changes requested):

server/game/cards/02.2-DD/BairdsBuildLoan.js
AaronSchnabel: I figured it would be easy enough for someone in that situation to add ghost rock to their stash and remove it manually from whatever other source they have available, but if you prefer I remove the condition, I will.

mmeldo: It is not difficult, but the other way is somewhat easier, and I plan to implement some kind of "redo", or "unpaying costs" in the future.

mmeldo (approved):

PR #465, opened 2021-05-12
use helper method to discard all cards at once.

### LINKED ISSUE #464: Use helper function for discarding cards
FYI, there is helper function `player.discardCards`

_Originally posted by @mmeldo in https://github.com/townteki/townsquare/pull/450#discussion_r629884545_

stopfstedt:
fixes #464

### DIFF
--- a/server/game/cards/01-DTR/AbramGrothe.js
+++ b/server/game/cards/01-DTR/AbramGrothe.js
@@ -16,7 +16,7 @@ class AbramGrothe extends DudeCard {
             onSuccess: (job, context) => {
                 const dudes = this.game.getDudesAtLocation(job.mark.gamelocation).filter(dude => dude.isWanted() || dude.hasKeyword('abomination'));
                 if(dudes.length) {
-                    context.player.discardCards(dudes);
+                    context.player.discardCards(dudes, false, () => true, {}, context);
                 }
             }
         });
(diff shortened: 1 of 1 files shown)


server/game/cards/01-DTR/AbramGrothe.js
```diff
@@ -16,7 +15,9 @@ class AbramGrothe extends DudeCard {
                 this.game.addMessage('{0} uses {1} to lead a job to discard all Abominations and wanted dudes at {2}', context.player, this, context.target),
             onSuccess: (job, context) => {
                 const dudes = this.game.getDudesAtLocation(job.mark.gamelocation).filter(dude => dude.isWanted() || dude.hasKeyword('abomination'));
-                dudes.forEach(dude => this.game.resolveGameAction(GameActions.discardCard({ card: dude }), context));
+                if(dudes.length) {
+                    context.player.discardCards(dudes);
```
mmeldo, line 19: Context should be passed to the function.
```suggestion
                    context.player.discardCards(dudes, false, () = > true, {}, context);
```
I know it does not look very nice, but it is needed in case engine has to decide if the discard can be done based on play type, or what player doing the discard.

mmeldo (changes requested):

server/game/cards/01-DTR/AbramGrothe.js
stopfstedt: > I know it does not look very nice, but it is needed in case engine has to decide if the discard can be done based on play type, or what player doing the discard.

ah, ok - thanks. i was wondering about those extra arguments.

stopfstedt: handled :heavy_check_mark:

mmeldo (approved):

PR #466, opened 2021-05-12
implements Make the Smart Choice

### LINKED ISSUE #175: DTR - Make the Smart Choice

stopfstedt:
fixes #175

### DIFF
--- a/server/game/cards/01-DTR/MakeTheSmartChoice.js
+++ b/server/game/cards/01-DTR/MakeTheSmartChoice.js
@@ -13,14 +13,31 @@ class MakeTheSmartChoice extends ActionCard {
             },
             handler: context => {
                 // "Reduce a dude's bullets by their influence."
-                this.applyAbilityEffect(context.ability, ability => ({
-                    match: context.target,
-                    effect: ability.effects.modifyBullets(-1 * context.target.influence)
-                }));
+                const influence = context.target.influence;
+                if(influence) {
+                    this.applyAbilityEffect(context.ability, ability => ({
+                        match: context.target,
+                        effect: ability.effects.modifyBullets(-1 * influence)
+                    }));
+                    this.game.addMessage(
+                        '{0} uses {1} to reduce {2}\'s bullets by their influence ({3})',
+                        context.target.controller,
+                        this,
+                        context.target,
+                        influence
+                    );
+                }
                 // "The dude's controller may move them home booted."
                 context.game.promptForYesNo(context.target.controller, {
                     title: 'Send home booted?',
-                    onYes: () => this.game.shootout.sendHome(context.target, context)
+                    onYes: () => {
+                        this.game.shootout.sendHome(context.target, context)
+                            .thenExecute(() => this.game.addMessage(
+                                '{0} uses {1} to send {2} home booted', 
+                                context.target.controller, 
+                                this, context.target
+                            ));
+                    }
                 });
             }
         });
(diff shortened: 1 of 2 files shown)


mmeldo (approved):

mmeldo (changes requested):
Messages missing.

stopfstedt:
> Messages missing.
good point @mmeldo, but please help me out here with the messaging.
should it be something to the extend of `X makes the smart choice and send Y home booted`, or do should this produce output on the bullet reduction as well?

mmeldo:
@stopfstedt 
For the first cards, I tried to include everything in one message, but sometimes it was too cumbersome with too many conditions. I would suggest to use 2 messages. One general to inform the action was executed:
`player1 uses X to reduce Y's bullets by Z` or `their influence (Z)`

The other when the other player decides to send him home:
`player2 makes the smart choice and sends Y home`

stopfstedt:
> player1 uses X to reduce Y's bullets by their influence (Z)
> player1 uses X to reduce Y's bullets by Z

added. the bullet reduction message only shows if the targeted dude actually has influence.

server/game/cards/01-DTR/MakeTheSmartChoice.js
```diff
@@ -0,0 +1,49 @@
...
+                // "Reduce a dude's bullets by their influence."
+                const influence = context.target.influence;
+                if(influence) {
+                    this.applyAbilityEffect(context.ability, ability => ({
+                        match: context.target,
+                        effect: ability.effects.modifyBullets(-1 * influence)
+                    }));
+                    this.game.addMessage(
+                        '{0} uses {1} to reduce {2}\'s bullets by their influence ({3})',
+                        context.target.controller,
```
mmeldo, line 24: Incorrect player using the ability.
```suggestion
                        context.player,
```

server/game/cards/01-DTR/MakeTheSmartChoice.js
```diff
@@ -0,0 +1,49 @@
...
+                        influence
+                    );
+                }
+                // "The dude's controller may move them home booted."
+                context.game.promptForYesNo(context.target.controller, {
+                    title: 'Send home booted?',
+                    onYes: () => {
+                        this.game.shootout.sendHome(context.target, context)
+                            .thenExecute(() => this.game.addMessage(
+                                '{0} uses {1} to send {2} home booted',
```
mmeldo, line 36: I would like to use the `X uses Y` if X is the one triggering the ability, while in this case he is forced to use it.
My suggestion would be something like 
`X sends Y home because he Z`
or
`X decides to send Y home because of the Z`

with _Z_ being the Make the Smart Choice card. Or some combination.

mmeldo (changes requested):

server/game/cards/01-DTR/MakeTheSmartChoice.js
stopfstedt: I see what you're getting at, but that makes for improper English. My suggestion here is to template the output message as `X makes the smart choice to send Y home booted.`

stopfstedt: image

stopfstedt: I see where you're going with the `because of Z`, but that's a bit awkward in English.
Suggesting we stick with `X  uses Y to send Z home` message template here.

image

stopfstedt: another option could be `X makes the smart choice and sends Y home.` that would be grammatically correct English and natural language.

stopfstedt: final idea on this: `X decides to Y and send Z home`, for example: `Player1 decides to Make the Smart Choice and sends Bobo home`.

mmeldo: 👍 That is basically what I had in mind, but obviously mi english is not that great. And I realize that many `traitReaction` abilities are not worded correctly because I used the same "style".

mmeldo (approved):

PR #468, opened 2021-05-13
Implemented Huntsmens Society

AaronSchnabel:


server/game/cards/04.2-NTB/HunstmensSociety.js
```diff
@@ -0,0 +1,18 @@
+const DeedCard = require('../../deedcard.js');
+
+class HunstmensSociety extends DeedCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            targetController: 'current',
+            condition: () => true,
+            match: card => card.getType() === 'dude' &&
+                (card.locationCard === this || card.isAdjacent(this.gamelocation)) &&
```
mmeldo, line 9: Just a very small thing, but I would rather use `gamelocation` to check if it is in the same location. Difference is that gamelocation is just a static `uuid` attribute of the location so no additional processing is done. But the `locationCard` getter searches thru players locations to find it, and then retrieves the card.
```suggestion
                (card.gamelocation === this.gamelocation || card.isAdjacent(this.gamelocation)) &&
```

mmeldo (changes requested):

mmeldo (approved):

PR #488, opened 2021-05-18
Implemented Ol' Fashioned Hangin'

AaronSchnabel:


server/game/cards/04.2-NTB/OlFashionedHangin.js
```diff
@@ -0,0 +1,29 @@
...
+                activePromptTitle: 'Choose a wanted dude to hang',
+                cardCondition: { location: 'play area', wanted: true, condition: card => card.bounty >= 2 },
+                cardType: ['dude']
+            },
+            message: context => 
+                this.game.addMessage('{0} uses {1} to lead a job to hang {2}', context.player, this, context.target),
+            onSuccess: (job, context) => {
+                this.game.resolveGameAction(GameActions.aceCard({ card: job.mark }), context);
+                if(job.leader.location === 'play area') {
+                    this.game.addMessage('{0} gains 1 CP', job.leader);
```
mmeldo, line 19: I would always start with the player and put more info here, for example:
`{0}\'s lead executioner {1} gains 1 CP as result of {2}`

server/game/cards/04.2-NTB/OlFashionedHangin.js
```diff
@@ -0,0 +1,29 @@
...
+            playType: 'noon',
+            target: {
+                activePromptTitle: 'Choose a wanted dude to hang',
+                cardCondition: { location: 'play area', wanted: true, condition: card => card.bounty >= 2 },
+                cardType: ['dude']
+            },
+            message: context => 
+                this.game.addMessage('{0} uses {1} to lead a job to hang {2}', context.player, this, context.target),
+            onSuccess: (job, context) => {
+                this.game.resolveGameAction(GameActions.aceCard({ card: job.mark }), context);
```
mmeldo, line 17: We also should check if the mark is in play before we ace it.
From the rulebook (Step 5. Is the Job Done?):
> Now resolve the effects of a successful job as given on the card, assuming the mark is still in play. If not, the job succeeds anyway but the mark is unaffected.

mmeldo (changes requested):

mmeldo (approved):

PR #493, opened 2021-05-19
Implemented Francisco Rosales

AaronSchnabel:


server/game/cards/08-GT/FranciscoRosales.js
```diff
@@ -0,0 +1,40 @@
+const DudeCard = require('../../dudecard.js');
+const GameActions = require('../../GameActions/index.js');
+const StandardActions = require('../../PlayActions/StandardActions.js');
+
+class FranciscoRosales extends DudeCard {
+    setupCardAbilities(ability) {
+        this.job({
+            title: 'Francisco Rosales',
+            playType: 'noon',
+            cost: ability.costs.bootLeader(),
```
mmeldo, line 10: This is not very intuitive, but in this case we should use `bootSelf()` cost because leader of the job is dude with the job ability. The `bootLeader()` cost is used only on other cards where player have to select a job leader.
```suggestion
            cost: ability.costs.bootSelf(),
```

server/game/cards/08-GT/FranciscoRosales.js
```diff
@@ -0,0 +1,40 @@
...
+            target: 'townsquare',
+            message: context =>
+                this.game.addMessage('{0} uses {1} to mark town square for a job', context.player, this),
+            onSuccess: (job, context) => {
+                this.game.resolveGameAction(
+                    GameActions.search({
+                        title: 'Select a horse or sidekick for Francisco',
```
mmeldo, line 17: No need to mention Francisco, his name will be in the title of the prompt.
```suggestion
                        title: 'Select a horse or sidekick',
```

mmeldo (changes requested):

mmeldo (approved):

PR #494, opened 2021-05-19
Card: Owl's Insight

AaronSchnabel:
The card works, but if you select multiple attachments at once, there is no indication of which one you are attaching, so you have to remember which order you selected them in. Once you've selected a good or spell, it prompts you again. If you select them one at a time, it works ok, but I'm not sure how to either force the goods/spell selection and attachment to happen one at a time and re-prompt or to allow multiselect and have the attachment prompt state the name of the card you're attaching.


mmeldo:
While implementing, discovered a bug: #496

mmeldo (approved):

PR #614, opened 2021-06-07
Card: Jasper Stone, Raven

mmeldo:
This is a tricky one so please test is much as you can.
Also fixed Wendy's Teethkickers and updated shootout posse effects.

### DIFF
--- a/docs/cardpool-status.md
+++ b/docs/cardpool-status.md
@@ -1,6 +1,6 @@
 # Card implementation status
 
-![Cards implemented](https://progress-bar.dev/285/?scale=875&title=implemented%20&width=200&suffix=%20cards) out of 875
+![Cards implemented](https://progress-bar.dev/287/?scale=875&title=implemented%20&width=200&suffix=%20cards) out of 875
 
 **Legend**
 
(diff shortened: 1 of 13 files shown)


stopfstedt (changes requested):
Jasper Stone checks out for me. 

However, I encountered a few issues with Raven:

1. Raven triggers twice in a round, not sure if that is correct tho. Once for winning the shootout on the location (+1 permanent bullet bonus), and then again during Sundown (+1 permanent bullet bonus). I don't think think that is correct, but will inquire in the discord.

raven

2. Raven prompts for a target to give a permanent bullet bonus, even if there is no eligible dudes in play (for example, all of my dudes already have a permanent bullet bonus). All of my dudes are highlighted, if i chose on with a permanent bullet bonus nothing happens and the effect fizzles (no additional bonus is granted), which is at least correct.

mmeldo:
PR updated with fixes for reported issues.

stopfstedt:
rules forum thread on Raven's effect: http://forums.pineboxentertainment.com/t/can-ravens-ability-grant-a-permanent-bullet-bonus-twice-per-round/4284

stopfstedt (approved):
LGTM.

PR #615, opened 2021-06-07
Implemented Xui Yin Chen

AaronSchnabel:


server/game/cards/05-IOUF/XuiYinChen.js
```diff
@@ -0,0 +1,27 @@
...
+class XuiYinChen extends DudeCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Raise posse influence and make them studs',
+            playType: 'shootout',
+            cost: ability.costs.bootSelf(),
+            message: context =>
+                this.game.addMessage('{0} uses {1}\'s ability on their posse to give them +1 influence and make them studs', context.player, this),
+            gameAction: 'increaseInfluence',
+            handler: (context) => {
```
mmeldo, line 12: The `gameAction` is not needed here. It is required only when you are selecting targets (to let engine know what we want to do with the target so it can prevent it if needed), or when used inside `actionContext` in case card is affecting itself or its parent.
```suggestion
                this.game.addMessage('{0} uses {1}\'s ability on their posse to give them +1 influence and make them studs', context.player, this),
            handler: (context) => {
```

mmeldo (changes requested):

mmeldo (approved):

PR #620, opened 2021-06-14
Card: Fallen Star, Land Purchase, Fannin' the Hammer

mmeldo:


mmeldo:
Please also test some of the changed cards:
- Fannin' the Hammer
- Run 'Em Down
- Mugging
- Hydro-puncher
- one of the techniques

stopfstedt:
*Land Purchase:* After the job succeeds, Land purchase does not end up in the discard pile. It vanishes from the game.

mmeldo:
> fallen star lets you chose an already booted dude as replacement target. that should not be the case, you cannot boot a dude that's already booted

This is one thing I was not so sure about. According to rules (or at least new rules regarding the requirements), only required targets are those that are formulated as: `Boot X to do Y`, or `Choose X to do Y` and so on.
I guess I will have to check with the rules guys.

mmeldo:
> After the job succeeds, Land purchase does not end up in the discard pile. It vanishes from the game. Otherwise, this card works fine

Haha, did not notice at all it disappeared 😄

mmeldo:
Answer is that Fallen Star redirects are required to boot since older cards still has the first sentence requirement. Another thing is that Land Purchase should not be able to choose itself from the discard pile 😄 Fixes pushed.

stopfstedt (changes requested):
kung fu combo is broken, the rest seems to work now.

mmeldo:
The combo problem was fixed on the master, but the fallen star branch was not updated so that is why it does not work. I have updated the branch so it should be also fixed here. It was a different problem.

stopfstedt (approved):
LGTM

PR #631, opened 2021-06-17
Card: Wang Men Wu, I'm Your Huckleberry

mmeldo:
Wang Men Wu's trait is very complicated, and I would say impossible to implement so it will be used generally in technique combos. Anyway, there is actually only one technique implemented (Rabbit's Lunar Leap) that is affected, and another one that is not yet implemented (Great Sage Matching Heaven). Thus I decided to implement Wu's trait directly into the combo condition. It is not ideal, but I am not aware of any other way.


mmeldo:
Wang Men Wu tested during play with m1, I'm Your Huckleberry also looks fine, going to merge this.

PR #679, opened 2021-07-04
implements As Nature and God Intended.

### LINKED ISSUE #608: W2D - As Nature and God Intended

stopfstedt:
fixes #608

### DIFF
--- a/server/game/cards/15-W2D/AsNatureAndGodIntended.js
+++ b/server/game/cards/15-W2D/AsNatureAndGodIntended.js
@@ -6,7 +6,7 @@ class AsNatureAndGodIntended extends ActionCard {
             title: 'As Nature and God Intended',
             playType: ['shootout'],
             message: context =>
-                this.game.addMessage('{0} playes {1}.', context.player, this), // TODO: Not sure about what to output here, regurgitating the card's effect is too verbose. [ST 2021/07/07]
+                this.game.addMessage(' {0} uses {1} to disable all Weapons and make all Kung Fu dudes studs until end of this round', context.player, this), // TODO: Not sure about what to output here, regurgitating the card's effect is too verbose. [ST 2021/07/07]
             handler: context => {
                 // All Weapons lose their bullet bonuses, traits, and Shootout abilities until the end of this round of the shootout.
                 this.lastingEffect(context.ability, ability => ({
--- a/server/game/cards/15-W2D/AsNatureAndGodIntended.js
+++ b/server/game/cards/15-W2D/AsNatureAndGodIntended.js
@@ -6,7 +6,7 @@ class AsNatureAndGodIntended extends ActionCard {
             title: 'As Nature and God Intended',
             playType: ['shootout'],
             message: context =>
-                this.game.addMessage(' {0} uses {1} to disable all Weapons and make all Kung Fu dudes studs until end of this round', context.player, this), // TODO: Not sure about what to output here, regurgitating the card's effect is too verbose. [ST 2021/07/07]
+                this.game.addMessage(' {0} uses {1} to disable all Weapons and make all Kung Fu dudes studs until end of this shootout round', context.player, this),
             handler: context => {
                 // All Weapons lose their bullet bonuses, traits, and Shootout abilities until the end of this round of the shootout.
                 this.lastingEffect(context.ability, ability => ({
--- a/server/game/cards/15-W2D/AsNatureAndGodIntended.js
+++ b/server/game/cards/15-W2D/AsNatureAndGodIntended.js
@@ -6,8 +6,7 @@ class AsNatureAndGodIntended extends ActionCard {
             title: 'As Nature and God Intended',
             playType: ['shootout'],
             message: context =>
-                this.game.addMessage('{0} playes {1}.', 
-                    context.player, this),
+                this.game.addMessage('{0} playes {1}.', context.player, this), // TODO: Not sure about what to output here, regurgitating the card's effect is too verbose. [ST 2021/07/07]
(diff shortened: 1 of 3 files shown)


stopfstedt:
@mmeldo i have one question with this card, and one implementation problem.

_question:_ can this effect target kung-fu dudes and weapons that are not part of the shootout?  not sure what the scope here is exactly.

_problem:_ the effects only seem to be taking hold on weapons and dudes controlled by the player who played this event. i can't quite figure out how to target the opponent's cards. any suggestions?

mmeldo:
> _question:_ can this effect target kung-fu dudes and weapons that are not part of the shootout?  not sure what the scope here is exactly.

I am almost 100% sure that shootout effects affect only dudes or goods attached to dudes participating in a shootout.

mmeldo:
> _problem:_ the effects only seem to be taking hold on weapons and dudes controlled by the player who played this event. i can't quite figure out how to target the opponent's cards. any suggestions?

Use `targetController: 'any'` in the `lastingEffect`

stopfstedt:
@mmeldo - i think this is _almost_ ready to go, the only thing that i can't quite figure out is how to turn off only 'shootout' abilities on attachments. `cannotTriggerCardAbilities()` seems to be the right method, i see the hook for passing in a callback as well, but i'm stuck at that point since i can't quite suss out which attribute of the effect to filter on. or perhaps this needs a dedicated method?

mmeldo:
> @mmeldo - i think this is _almost_ ready to go, the only thing that i can't quite figure out is how to turn off only `shootout` abilities on attachments. `cannotTriggerCardAbilities()` seems to be the right method, i see the hook for passing in a callback as well, but i'm stuck at that point since i can't quite suss out which attribute of the effect to filter on. or perhaps this needs a dedicated method? as it stands, this implementation turns off _all_ card abilities on attachments that are in the fight, including cheating resolutions, which is obviously not correct.

You can pass a condition to the `cannotTriggerCardAbilities()` effect with `ability` parameter.
Then you can check if the ability is a shootout ability by checking if it has "shootout" in the `playType`. There can be abilities with "noon/shootout" but since the  `As Nature and God Intended` will last only in shootout we do not care. And I am not aware of any abilities that can be triggered in shootout plays and also in resolution.
```
ability => ability.playType && ability.playType.includes('shootout')
```
I did not test this, but I believe that should work.

stopfstedt:
@mmeldo thanks for the feedback, i'll try to wrap this one up today and get started with the next one.

stopfstedt:
> Edit: You can actually make sure you are currently in "shootout plays" but adding this condition:
> `ability.game.isShootoutPlayWindow()`

this didn't work, and doesn't seem to be necessary. worth a try though.

I found a different problem though, and that has to do with the card blanking effect of `blankExcludingKeywords`. this is too far- reaching, as this flags the attachment as (partially) blanked and therefore  blocks other, non-shootout abilities (like cheatin' resolutions) as well.
For example, this turns off Quickdraw Handgun's cheatin' resolution. 

What would be needed here is a more fine-grained approach to handling blanking effects/blank card states. I peeked at the relevant code and separating the blanking of traits from the blanking of abilities seems a bit out of scope for this card implementation/would require a more involved rewrite.

I'll file this as a separate issue and sit on this card for the time being.

mmeldo:
Added new effects: `traitBlank` and `bulletBonusBlank`

stopfstedt:
@mmeldo happy to report that the new blanking effects work like a charm. Click-tested  with a full assortment of weapon attachment - Shotgun, Flame-Thrower, Quickdraw Handgun, etc., this now works as intended.

If you have any thoughts on how to improve the message output for this action please let me know (or add one yourself), i left this as a TODO in there.

mmeldo:
> If you have any thoughts on how to improve the message output for this action please let me know (or add one yourself), i left this as a TODO in there.

Well, I have 2 suggestions. One is verbose:
`{0} uses {1} to have all Weapons lose their traits, bullet bonuses and Shootout abilities, and make all Kung Fu dudes studs until end of this round`
and the other is not, but it is not that clear:
`{0} uses {1} to disable all Weapons and make all Kung Fu dudes studs until end of this round`

You can use any of these and I will be fine with that. Or neither and maybe you can come up with something better.

stopfstedt:
i went with the second suggestion, thanks for that. ready for re-review.

server/game/cards/15-W2D/AsNatureAndGodIntended.js
```diff
@@ -0,0 +1,43 @@
+const ActionCard = require('../../actioncard.js');
+
+class AsNatureAndGodIntended extends ActionCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'As Nature and God Intended',
+            playType: ['shootout'],
+            message: context =>
+                this.game.addMessage(' {0} uses {1} to disable all Weapons and make all Kung Fu dudes studs until end of this round', context.player, this), // TODO: Not sure about what to output here, regurgitating the card's effect is too verbose. [ST 2021/07/07]
```
mmeldo, line 9: You can delete the TODO and I know I suggested it, but I believe the message should end: "... until end of this shootout round" to make clear it does not last until the end of the day (round).

mmeldo (changes requested):

server/game/cards/15-W2D/AsNatureAndGodIntended.js
stopfstedt: :heavy_check_mark:

mmeldo:
Man, that was fast 🚄

mmeldo (approved):

PR #682, opened 2021-07-07
Trait and Bonuses blanking

### LINKED ISSUE #680: "blank only traits" ability effect needed.
see the text on "As Nature and God Intended"

> All Weapons lose their bullet bonuses, traits, and Shootout abilities until the end of this round of the shootout.

that effect should leave abilities other than shootout abilities, like cheatin' resolutions, intact.

however there is currently no way to partially blank cards described in this manner - we only have ways to _fully_ blank cards, and to blank cards excluding keywords.

what's needed is a mechanism that allows for blanking of traits only, card abilities can be selectively turned off by adding other ability effects.

this will require changes to `BaseCard` and `Cardaction::meetsRequirements()`, possibly other places as well.

mmeldo:
fixes #680

### DIFF
--- a/server/game/basecard.js
+++ b/server/game/basecard.js
@@ -228,7 +228,11 @@ class BaseCard {
             throw new Error(`'${location}' is not a supported effect location.`);
         }
 
-        this.abilities.persistentEffects.push(Object.assign({ duration: 'persistent', location: location }, properties));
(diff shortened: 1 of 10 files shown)


mmeldo:
This is required for #679

PR #687, opened 2021-07-15
Implemented Tawodi

AaronSchnabel:


mmeldo (commented):
Your code is fine, but I think we should do a general implementation for cost reduction minimum. Let's update the card once it is implemented.

mmeldo (approved):
The cost reduction minimum was implemented and card updated. Could you please test it along with Morgan Cattle Co. and Morgan Lash?

AaronSchnabel:
> The cost reduction minimum was implemented and card updated. Could you please test it along with Morgan Cattle Co. and Morgan Lash?

Everything seems to be working just fine.

PR #697, opened 2021-07-20
Card: Lorena Corbett, Mason Adler

mmeldo:
Added required effects and fixed:
 - bug where action was not marked as taken in case if condition failed
 - bug where +1 was added to the selected joker value


mmeldo:
@AaronSchnabel do you think you will have time to test these 2 cards?

AaronSchnabel:
> @AaronSchnabel do you think you will have time to test these 2 cards?

I have now tested them a bit and they seem to be working for me.

PR #699, opened 2021-07-21
Implemented Fort 51

AaronSchnabel:
I fixed a small typo when inventing gadgets so that it says "invents" instead of "invent" and I added the isAtHome() function to goodscards.js so that goodscards can check if they are at home or not. Not sure if that is the best approach of if I should have just added that functionality to the Fort 51 card instead.

### DIFF
--- a/server/game/cards/13-O4B/Fort51.js
+++ b/server/game/cards/13-O4B/Fort51.js
@@ -52,8 +52,8 @@ class Fort51 extends OutfitCard {
                     title: 'Do you want to discard a card?',
                     onYes: () => {
                         context.player.discardFromHand(1, discarded => {
-                            this.game.addMessage('{0} uses {1} to discard {2}', context.player, this, discarded)
-                        },   { title: this.title });
+                            this.game.addMessage('{0} uses {1} to discard {2}', context.player, this, discarded);
+                        }, { title: this.title });
                     }
                 });
             }
--- a/server/game/cards/13-O4B/Fort51.js
+++ b/server/game/cards/13-O4B/Fort51.js
@@ -15,8 +15,9 @@ class Fort51 extends OutfitCard {
                     cardType: 'dude',
                     gameAction: 'addBounty',
                     onSelect: (player, card) => {
-                        this.game.resolveGameAction(GameActions.addBounty({ card, amount: 1 }), context);
-                        this.game.addMessage('{0} gives {1} 1 bounty thanks to {2}', player, card, this);
+                        this.game.resolveGameAction(GameActions.addBounty({ card, amount: 1 }), context).thenExecute(() => {
+                            this.game.addMessage('{0} gives {1} 1 bounty thanks to {2}', player, card, this);
+                        });
                         return true;
                     }
                 });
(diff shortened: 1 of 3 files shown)


server/game/cards/13-O4B/Fort51.js
```diff
@@ -0,0 +1,66 @@
...
+                card.hasKeyword('gadget') && card.controller === this.controller && !card.isAtHome()
+            ),
+            target: {
+                activePromptTitle: 'Select a dude',
+                waitingPromptTitle: 'Waiting for opponent to select a dude',
+                cardCondition: { 
+                    location: 'play area', 
+                    condition: card =>
+                        card.controller !== this.controller && card.isWanted()
+                },
```
mmeldo, line 39: This works too, but since there are specific properties for these conditions, let's use them.
```suggestion
                cardCondition: { 
                    location: 'play area', 
                    controller: 'opponent',
                    wanted: true
                },
```

server/game/cards/13-O4B/Fort51.js
```diff
@@ -0,0 +1,66 @@
...
+            handler: context => {
+                this.game.promptForSelect(context.player, {
+                    activePromptTitle: 'Select a dude',
+                    waitingPromptTitle: 'Waiting for opponent to select dude',
+                    cardCondition: card => card.location === 'play area' && card.controller !== this.controller && card.influence < context.event.gadget.cost,
+                    cardType: 'dude',
+                    gameAction: 'addBounty',
+                    onSelect: (player, card) => {
+                        this.game.resolveGameAction(GameActions.addBounty({ card, amount: 1 }), context);
+                        this.game.addMessage('{0} gives {1} 1 bounty thanks to {2}', player, card, this);
```
mmeldo, line 19: Put the `addMessage` in `thenExecute` after the `resolveGameAction`.

server/game/cards/13-O4B/Fort51.js
```diff
@@ -0,0 +1,66 @@
...
+const OutfitCard = require('../../outfitcard.js');
+
+class Fort51 extends OutfitCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onGadgetInvented: event => event.gadget.hasKeyword('Weapon') || event.gadget.getType() === 'dude' && event.player === this.controller
+            },
+            handler: context => {
+                this.game.promptForSelect(context.player, {
```
mmeldo, line 11: This works too, but it does not trigger specific events for selecting other targets that are required for some reacts (or can be in the future). That is why we should use a specific function `selectAnotherTarget`.
```suggestion
                context.ability.selectAnotherTarget(context.player, context, {
```

mmeldo (changes requested):

mmeldo:
I think we can keep the `isAtHome()` function. I will need to refactor the `GameLocation` handling anyway as it can cause crashes in specific situations.

server/game/cards/13-O4B/Fort51.js
AaronSchnabel: Done.

AaronSchnabel:
I removed "the" from the inventing messages so now it will read "Player successfully invents gadget using Scientist" instead of "Player successfully invents gadget using the Scientist"

server/game/cards/13-O4B/Fort51.js
```diff
@@ -0,0 +1,67 @@
+const GameActions = require('../../GameActions/index.js');
+const OutfitCard = require('../../outfitcard.js');
+
+class Fort51 extends OutfitCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onGadgetInvented: event => event.gadget.hasKeyword('Weapon') || event.gadget.getType() === 'dude' && event.player === this.controller
```
mmeldo, line 8: The `and` (&&) operator has a priority so this condition would not check the controller in case the gadget has 'weapon' keyword, and will trigger also if opponent invents weapon gadget.
```suggestion
                onGadgetInvented: event => (event.gadget.hasKeyword('Weapon') || event.gadget.getType() === 'dude') && event.player === this.controller
```

mmeldo (changes requested):

mmeldo (approved):

PR #703, opened 2021-07-22
Implemented Auntie Sinister

AaronSchnabel:


server/game/cards/13-O4B/AuntieSinister.js
```diff
@@ -0,0 +1,27 @@
...
+            condition: () => this.game.shootout && this.isParticipating() && this.hasAttachmentWithKeywords('mystical'),
+            match: this,
+            effect: ability.effects.dynamicSkillRating('huckster', () => this.influence)
+        });
+
+        this.reaction({
+            when: {
+                onPullSuccess: event => 
+                    event.pullingDude === this && event.pullingDude.isParticipating() && 
+                        ((event.pulledValue + this.getSkillRating('huckster')) - 6 >= event.source.difficulty) && event.source.isHex()
```
mmeldo, line 15: Use `totalPullValue` from the event context as there can be other bonuses.
Also, just to be sure add the type check for a spell. I know there is `isHex()` check, but if there will be a scenario that Auntie will pull for some other reason (source), it can crash because `isHex` function is defined only for spells and not other card types.
```suggestion
                        (event.context.totalPullValue - 6 >= event.source.difficulty) && event.source.getType() === 'spell' && event.source.isHex()
```

mmeldo (changes requested):

mmeldo (approved):

PR #716, opened 2021-07-28
Implements Bayou Vermilion Railroad

### LINKED ISSUE #657: O4B - Bayou Vermilion Railroad

stopfstedt:
fixes #657

### DIFF
--- a/server/game/cards/13-O4B/BayouVermilionRailroad.js
+++ b/server/game/cards/13-O4B/BayouVermilionRailroad.js
@@ -1,4 +1,5 @@
 const OutfitCard = require('../../outfitcard.js');
+const GameActions = require('../../GameActions/index.js');
 
 class BayouVermilionRailroad extends OutfitCard {
     setupCardAbilities(ability) {
@@ -18,15 +19,17 @@ class BayouVermilionRailroad extends OutfitCard {
                 }
             },
             handler: context => {
+                this.game.resolveGameAction(GameActions.bootCard({ card: context.target }), context);
+                context.player.modifyGhostRock(1);
                 this.untilEndOfPhase(context.ability, ability => ({
                     condition: () => true,
-                    match: context.target.getGameLocation(),
+                    match: card => context.target.getGameLocation().uuid === card.uuid,
                     effect: [
-                        ability.effects.modifyProduction(0),
-                        ability.effect.modifyControl(0)
+                        ability.effects.setProduction(0),
+                        ability.effects.setControl(0)
                     ]
                 }), 'upkeep');
-                
+
             },
             source: this
         });
--- a/server/game/cards/13-O4B/BayouVermilionRailroad.js
+++ b/server/game/cards/13-O4B/BayouVermilionRailroad.js
@@ -15,8 +15,9 @@ class BayouVermilionRailroad extends OutfitCard {
                 activePromptTitle: 'Select a Mystical card',
                 cardCondition: { 
                     location: 'play area',
-                    condition: card => card.hasKeyword('mystical') && card.isAtDeed() && !card.booted && card.canBeBooted()
-                }
+                    condition: card => card.hasKeyword('mystical') && card.isAtDeed()
+                },
+                gameAction: 'boot'
             },
             handler: context => {
                 this.game.resolveGameAction(GameActions.bootCard({ card: context.target }), context);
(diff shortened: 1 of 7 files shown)


stopfstedt:
@mmeldo this seems to work now, for the most part. 

the remaining issue is that the "untilEndOfPhase" effect doesn't stick once the _Mystical_ cards is moved off of the location that got nerfed. 

Example: 
1. My dude with Evanor moves to my opponent's Railway Station
2. I use the Bayou Vermilion Railroad's noon ability with Evanor to tank the Railway Station's CP and production.
3. My dude moves elsewhere - the railway station now regains its original CP and prod.

I need to make the initial location (Railway Station) stick somehow. Any suggestions?

mmeldo:
> @mmeldo this seems to work now, for the most part.
> 
> the remaining issue is that the "untilEndOfPhase" effect doesn't stick once the _Mystical_ cards is moved off of the location that got nerfed.
> 
> Example:
> 
> 1. My dude with Evanor moves to my opponent's Railway Station
> 2. I use the Bayou Vermilion Railroad's noon ability with Evanor to tank the Railway Station's CP and production.
> 3. My dude moves elsewhere - the railway station now regains its original CP and prod.
> 
> I need to make the initial location (Railway Station) stick somehow. Any suggestions?

The effects can be state dependent or not. State dependent effects get reapplied after every event, which sometime (as in your case) is something you do not want. To make a state dependent effect, there are 2 ways:
- fixed: `isStateDependent: true` property defined for the effect (e.g. any dynamic stat, for example `dynamicBullets`)
- by specifying `condition` along with `match` being a function

What you have is the second one which means the effect is state dependent. If you want to have it applied only once and be done, remove the `condition` and pass a card or array of cards in the `match` instead of the function. So in our case, it should be enough to have `match: context.target.locationCard`

mmeldo:
I did a small refactor regarding the location functions, I hope you do not mind. The problem here is that using the `getGameLocation()` is dangerous as it can return `null` and cause crashes. Thus it is better to put it into the helper functions with proper checks.

server/game/cards/13-O4B/BayouVermilionRailroad.js
```diff
@@ -0,0 +1,39 @@
...
+            title: 'Bayou Vermilion Railroad',
+            playType: 'noon',
+            cost: ability.costs.bootSelf(),        
+            target: {
+                activePromptTitle: 'Select a Mystical card',
+                cardCondition: { 
+                    location: 'play area',
+                    condition: card => card.hasKeyword('mystical') && card.isAtDeed() && !card.booted && card.canBeBooted()
+                }
+            },
```
mmeldo, line 20: It is cleaner to use `gameAction` as that is used in the engine to check if the action is allowed and some other validation.
```suggestion
            target: {
                activePromptTitle: 'Select a Mystical card',
                cardCondition: { 
                    location: 'play area',
                    condition: card => card.hasKeyword('mystical') && card.isAtDeed()
                },
                gameAction: 'boot'
            },
```

server/game/cards/13-O4B/BayouVermilionRailroad.js
```diff
@@ -0,0 +1,39 @@
...
+            handler: context => {
+                this.game.resolveGameAction(GameActions.bootCard({ card: context.target }), context);
+                context.player.modifyGhostRock(1);
+                this.untilEndOfPhase(context.ability, ability => ({
+                    match: context.target.locationCard,
+                    effect: [
+                        ability.effects.setProduction(0),
+                        ability.effects.setControl(0)
+                    ]
+                }), 'upkeep');
```
mmeldo, line 30: > you may reduce the deed's control points and production to 0 until after the next Upkeep phase

It says you may reduce, so we should put a prompt to ask player if he wants to reduce.

mmeldo (changes requested):

server/game/cards/13-O4B/BayouVermilionRailroad.js
stopfstedt: good catch, i'll wrap it into a prompt tonight.

stopfstedt: I added the prompt and some game messages.

mmeldo (approved):

Issue #718, opened 2021-07-28
Refactor shootout restriction abilities


mmeldo:
Shootout restriction abilities are not really needed and their implementation is not that good anyway. 
The part that needs taking care of is `cannotBringDudesIntoPosseByShootout` effect in `ScoopHound.js` and `MasonAdler.js`.

PR #719, opened 2021-07-28
Card: Johnny Ringo, Sgt. Elijah Clay

mmeldo:
Plus fix for PointBlank.js and updates on the shootout effects.


mmeldo:
Partly removed restrictions for shootout. Rest in #718

PR #805, opened 2021-08-12
implements Grimoires & More

### LINKED ISSUE #801: TSs - Grimoires & More

stopfstedt:
fixes #801

### DIFF
--- a/server/game/cards/09.3-TSs/GrimoiresAndMore.js
+++ b/server/game/cards/09.3-TSs/GrimoiresAndMore.js
@@ -1,0 +1,50 @@
+const DeedCard = require('../../deedcard.js');
+const GameActions = require('../../GameActions/index.js');
+const StandardActions = require('../../PlayActions/StandardActions.js');
+
+class GrimoiresAndMore extends DeedCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            title: 'React: Grimoires & More',
+            when: {
+                onCardEntersPlay: event => event.card.controller === this.controller && event.card.getType() === 'dude' && event.card.hasKeyword('huckster')
+            },
+            cost: ability.costs.bootSelf(),
+            handler: context => {
+                const dude = context.event.card;
+                context.player.pull((pulledCard) => {
+                    if('clubs' === pulledCard.suit.toLowerCase()) {
+                        this.game.resolveGameAction(GameActions.sendHome({ card: dude, needToBoot: true}), context);
+                    } else {
+                        this.game.resolveGameAction(
+                            GameActions.search({
+                                title: 'Select a Hex to attach',
+                                match: { keyword: 'hex', type: 'spell' },
+                                location: ['discard pile'],
+                                numToSelect: 1,
+                                message: {
+                                    format: '{player} plays {source} to search their discard pile, and puts {searchTarget} into play'
+                                },
+                                cancelMessage: {
+                                    format: '{player} plays {source} to search their discard pile, but does not find a Hex'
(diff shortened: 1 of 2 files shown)


server/game/cards/09.3-TSs/GrimoiresAndMore.js
```diff
@@ -0,0 +1,50 @@
...
+        this.reaction({
+            title: 'React: Grimoires & More',
+            when: {
+                onCardEntersPlay: event => event.card.controller === this.controller && event.card.getType() === 'dude' && event.card.hasKeyword('huckster')
+            },
+            cost: ability.costs.bootSelf(),
+            handler: context => {
+                const dude = context.event.card;
+                context.player.pull((pulledCard) => {
+                    if('clubs' === pulledCard.suit.toLowerCase()) {
```
mmeldo, line 16: I would rather use `pulledSuit` parameter from the pull function. That is because if there will be in the future some card that would replace the pull effect and maybe change suit, the new suit will be passed as `pulledSuit` while the `pulledCard` will have the default suit. (e.g. `AnalyticalCognisizer` for gadgets)
```suggestion
                context.player.pull((pulledCard, pulledValue, pulledSuit) => {
                    if('clubs' === pulledSuit.toLowerCase()) {
```

mmeldo (changes requested):

server/game/cards/09.3-TSs/GrimoiresAndMore.js
```diff
@@ -0,0 +1,50 @@
...
+            title: 'React: Grimoires & More',
+            when: {
+                onCardEntersPlay: event => event.card.controller === this.controller && event.card.getType() === 'dude' && event.card.hasKeyword('huckster')
+            },
+            cost: ability.costs.bootSelf(),
+            handler: context => {
+                const dude = context.event.card;
+                context.player.pull((pulledCard, pulledValue, pulledSuit) => {
+                    if('clubs' === pulledSuit.toLowerCase()) {
+                        this.game.resolveGameAction(GameActions.sendHome({ card: dude, needToBoot: true}), context);
```
mmeldo, line 17: Last thing (sorry, did not notice before). There also should be a message here that Grimoires has been used but huckster was sent home due to clubs pull.

mmeldo (changes requested):

server/game/cards/09.3-TSs/GrimoiresAndMore.js
stopfstedt: good catch, i'll add that message over lunch today.

stopfstedt: :heavy_check_mark:

mmeldo: Do not want to be a pain 😄 But that message seems like the deed was used to send the huckster home, and not that it has failed (pulled clubs) and was sent home because of that. I would format is similarly to other experimental cards:
> {0} uses {1} to pull {2} and has to send {3} home because it is a club
or something like that. What do you think?

stopfstedt: that's much better. will add it later tonight. thanks!

stopfstedt: :heavy_check_mark:

stopfstedt: ready for re-review.

mmeldo (approved):

PR #818, opened 2021-08-20
language fix.

stopfstedt:
minor language fix, i discovered this issue when i was pilfering Maggie's ability for Grimoires & More (see #805 ).


mmeldo:
Thanks. I suspect there will be quite a few of those.

PR #819, opened 2021-08-20
Card: QUATERMAN PRIME, John "Aces" Radcliffe

mmeldo:


mmeldo:
Tested also with the `Bayou Vermillion Railroad`, but the noon ability actually cannot be used as `QUATERMAN PRIME` cannot be booted. But that worked along with the discount.

PR #824, opened 2021-08-24
Card: Lacy O'Malley, Zeb Whateley-Dupont

mmeldo:
Updated `moveCard` player function (created a new one to take also context) so it can be checked if the moving card is done due to card ability.
Updated effects that uses minimal bullets (cannot be less than).

### DIFF
--- a/server/game/basecard.js
+++ b/server/game/basecard.js
@@ -50,6 +50,8 @@ class BaseCard {
         this.wealth = cardData.wealth;
         this.permanentBullets = 0;
         this.startingSize = 1;
+        this.maxBullets = null;
+        this.minBullets = 0;        
 
         this.tokens = {};
 
@@ -88,8 +90,8 @@ class BaseCard {
     }
 
     get bullets() {
-        if(this.currentBullets < 0) {
-            return 0;
+        if(this.currentBullets < this.minBullets) {
+            return this.minBullets;
         }
         return this.currentBullets;
     }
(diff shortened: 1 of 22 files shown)


mmeldo:
Help with testing needed:
- Maggie Harris
- Harold Aimslee
- Devil's Six Gun
or some other card that puts a card from discard pile to play.

stopfstedt:
Testing Maggie's Job - this works as intended. Job goes through, horse is selected but doesn't enter play. However, the message output could use some improvement. perhaps an additional `Lacy O'Malley prevents {0} from entering play.` would be helpful to the players here.

image

stopfstedt:
Devil's Six Gun works as intended - search for Joker is triggered, but doesn't enter hand. The "some effect prevents" message output is spot on, something like that should be added to Maggie's output.

image

stopfstedt:
Testing with Harold Aimslee - works as intended! 

image

stopfstedt (changes requested):
Needs additional message output if the target of Maggie Harris' job resolution was prevented from entering play. LGTM otherwise.

mmeldo:
> Needs additional message output if the target of Maggie Harris' job resolution was prevented from entering play. LGTM otherwise.

When updating the `PutIntoPlay` action to show message that the target (horse) was prevented from entering play, I found another problem. Problem was that the `PutIntoPlay` action could not be checked if it was a card ability so I fixed that.
Another thing is that there is a lot more cards similar to Maggie that needed updating. And a special chapter was `Dr. Dawn Edwards` vs `Eve Henry`. It was not correctly implemented before so I fixed it.

To test this, please test at least 3 other cards putting stuff into play from discard. For example Pete Spence, Land Purchase and Recruitment Drive along with Dr. Dawn Edwards and Eve Henry combo. I will test some other cards.

stopfstedt:
> To test this, please test at least 3 other cards putting stuff into play from discard. For example Pete Spence, Land Purchase and Recruitment Drive along with Dr. Dawn Edwards and Eve Henry combo. I will test some other cards.

re-testing tonight.

PR #827, opened 2021-08-24
Card: Spiritual Society, Property Is Theft

mmeldo:


mmeldo:
Needs ruling on Spiritual Society.

mmeldo:
Answer was something like:
> homes should do nice things. I'd say just have the card

So I will update to have the additional card draw on sundown even if there was no card drawn.

PR #861, opened 2021-08-30
Card: Hostile Takeover, Frontier Feud

mmeldo:
Plus small fix for Trading action - it now can happen only in a controlled location.

### DIFF
--- a/server/game/dudecard.js
+++ b/server/game/dudecard.js
@@ -234,7 +234,9 @@ class DudeCard extends DrawCard {
         this.action({
             title: 'Trade',
             abilitySourceType: 'game',
-            condition: () => this.game.currentPhase === 'high noon' && this.hasAttachmentForTrading(),
+            condition: () => this.game.currentPhase === 'high noon' && 
+                this.isInControlledLocation() &&
+                this.hasAttachmentForTrading(),
             target: {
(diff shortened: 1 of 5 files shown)


server/game/dudecard.js
```diff
@@ -234,7 +234,9 @@ class DudeCard extends DrawCard {
         this.action({
             title: 'Trade',
             abilitySourceType: 'game',
-            condition: () => this.game.currentPhase === 'high noon' && this.hasAttachmentForTrading(),
+            condition: () => this.game.currentPhase === 'high noon' && 
+                this.isInControlledLocation() &&
```
anthony-colpron, line 238: I'd suggest not putting mechanic fixes in the same PR where we add cards.

anthony-colpron:
@mmeldo `Frontier Feud` looks like it applies the rules properly.

server/game/dudecard.js
mmeldo: I basically agree with you but at this stage it just slows me down. Many times, I am working on a card and during testing find some small thing so just fix it there. To create new branch, separate the change, then new PR is in my opinion at this time unnecessary. What I try to do is describe the additional change in the commit message.
Once the code will be more stable, then I definitely want to make it separate.

PR #879, opened 2021-08-31
Card: Roberto Muratore, Adrian Vallejo

mmeldo:
During the testing I found out the handling of pulled cards in card scripts (such as experimental) is not done properly, or not for all cards. To resolve this, I added new event handler into the `pull` function in `player.js` to handle the pull card every time a context is passed. Therefore the only responsibility card script has is to pass the context.


mmeldo:
I will need help with testing this. As mentioned in the description, bigger change regarding pulled cards was introduced, so any cards doing pulls, e.g.: R&D Ranch, Elander Boldman, Grimoires & More etc.
Pay attention to the pulled card and if it is properly handled (discarded after the ability, or aced if it is joker).

PR #881, opened 2021-08-31
Card: Fate Dispenser, Ol'Howard

mmeldo:
Bigger change containing handling of Improvement cards and correctly setting controllers for attachments.


mmeldo:
In this PR, Improvements should be properly handled. Also controller for all attachments should be correctly set to the controller of parent, and Totems triggering set to anyone's Shaman.
Only Improvement implemented is in this PR (Fate Dispenser), Totems for testing: Mother Bear's rage or Spirit Trail

PR #941, opened 2021-09-10
Florentino value fix

AaronSchnabel:
Fixed the bug where Florentino's value would remain changed if he was discarded with bounty.


mmeldo:
This fix is fine for the symptom, but unfortunately does not deal with the root cause. There are 2 other problems I see:
 - for the trait to work Florentino still had to have bounty which should be cleared when he left play
 - persistent effect should be "dormant" when card is not in play, if not otherwise defined (e.g. Comin' Up Roses)

mmeldo:
This will be fixed properly in #1080

PR #945, opened 2021-09-13
Unpay for Shoppin'

mmeldo:
When player is doing a Shoppin' action, it can happen they do not have valid parent to place it on, or they just want to cancel before selecting the parent. This change adds an unpay for this case that will return the GR, prevent the order to be moved to next player and refreshes any reducers that were used.


mmeldo:
Needs to implement proper returning of GR (can be different source than player's stash)

PR #955, opened 2021-09-14
GT:Auto-Gatling

PesiTheWizard:
* Added file for Auto-Gatling

* Marking Auto-Gatling


mmeldo:
The implementation is good, but there is an annoying effect caused by the Gatling lowering influence. In case a 1 inf dude comes to a deed with Gatling, they take control but that is immediately taken back. Would be ok if it was just once, but from some reason, this happens 10 times spamming the chat log. Will have to take a look at that first.

mmeldo (approved):

PR #1128, opened 2021-10-13
Card: TLS - Tummy Twister

PesiTheWizard:
* Adding file for Tummy Twister

* Marking Tummy Twister


mmeldo:
This will need some updates in the game engine because moves of multiple dudes (e.g. during job) are done one by one in random order and it should be simultaneous.

mmeldo (approved):

PR #1169, opened 2021-10-26
Implement React timer plus A Slight Modification

### LINKED ISSUE #623: Remove cancel timer
`server/game/gamesteps/CancelTimer.js`
used in `TriggeredAbilityWindow`

mmeldo:
Reusing Cancel timer from AGOT as a React timer for React abilities. This will be especially useful for actions. To cover if player has action with React in hand, the timer will prompt even if the action is not in hand.
Plus refactor to remove unused settings.
New Timed Reaction Window:
image

If "Show timer if actions with React in deck" is checked, then there is timed prompt anytime the event triggering the action's react occurs .. even if the action card is not in hand. This way opponent does not know if you really have the action card in hand.
A Slight Modification was implemented along with it, so that is best for testing, but also Get Off My Land! or Takin' Ya With Me.

UI part: townteki/townsquare-client#88
closes #623


stopfstedt (approved):
click-tested, LGTM.

PR #1170, opened 2021-10-27
Card: 2T2D - Bedazzle

### LINKED ISSUE #571: 2T2D - Bedazzle

PesiTheWizard:
* Adding file for Bedazzle

* Marking Bedazzle

_(Probably fine)_

closes #571


server/game/cards/12-2T2D/Bedazzle.js
```diff
@@ -0,0 +1,38 @@
...
+            playType: 'cheatin resolution',
+            cost: ability.costs.bootSelf(),
+            difficulty: 3,
+            choosePlayer: true,
+            onSuccess: (context) => {
+                context.chosenPlayer.modifyRank(-2, context);
+                this.game.addMessage('{0} uses {1} to lower {2}\'s draw hand by 2 ranks; Current rank is {3}',
+                    context.player, this, context.chosenPlayer, context.chosenPlayer.getTotalRank());
+                if(this.game.shootout) {
+                    this.game.promptForSelect(context.player, {
```
mmeldo, line 16: In case you are selecting another target in the handler, you should use `context.ability.selectAnotherTarget`.
```suggestion
                    context.ability.selectAnotherTarget(context.player, context, {

```

server/game/cards/12-2T2D/Bedazzle.js
```diff
@@ -0,0 +1,38 @@
...
+                        waitingPromptTitle: 'Waiting for opponent to select dude',
+                        cardCondition: card => card.location === 'play area' &&
+                            card.controller === context.player.getOpponent() &&
+                            card.isParticipating(),
+                        cardType: 'dude',
+                        onSelect: (p, dazzlee) => {
+                            const dazzleAmount = Math.max(dazzlee.bullets, 0);
+                            ability.effects.modifyBullets(0 - dazzleAmount);
+                            this.game.addMessage('{0} gets -{1} bullets due to the {2}', dazzlee, dazzleAmount, this);
+                        }
```
mmeldo, line 27: Add `source: this` so the waiting prompt is automatically generated based on the card. I know it is not used everywhere, just realized it lately so would like to add it to new cards.
So the waiting prompt line should be deleted.
```suggestion
                        },
                        source: this
```

mmeldo (changes requested):

server/game/cards/12-2T2D/Bedazzle.js
```diff
@@ -0,0 +1,40 @@
...
+                    context.ability.selectAnotherTarget(context.player, context, {
+
+                        activePromptTitle: 'Choose opposing dude to bedazzle',
+                        waitingPromptTitle: 'Waiting for opponent to select dude',
+                        cardCondition: card => card.location === 'play area' &&
+                            card.controller === context.player.getOpponent() &&
+                            card.isParticipating(),
+                        cardType: 'dude',
+                        onSelect: (p, dazzlee) => {
+                            const dazzleAmount = Math.max(dazzlee.bullets, 0);
```
mmeldo, line 25: This is not needed as basecard has a `bullets` getter that always returns number higher than 0.

server/game/cards/12-2T2D/Bedazzle.js
```diff
@@ -0,0 +1,40 @@
...
+                            ability.effects.modifyBullets(0 - dazzleAmount);
+                            this.game.addMessage('{0} gets -{1} bullets due to the {2}', dazzlee, dazzleAmount, this);
```
mmeldo, line 27: We have an internal rule that message should always start with a user.

server/game/cards/12-2T2D/Bedazzle.js
```diff
@@ -0,0 +1,40 @@
...
+            choosePlayer: true,
+            onSuccess: (context) => {
+                context.chosenPlayer.modifyRank(-2, context);
+                this.game.addMessage('{0} uses {1} to lower {2}\'s draw hand by 2 ranks; Current rank is {3}',
+                    context.player, this, context.chosenPlayer, context.chosenPlayer.getTotalRank());
+                if(this.game.shootout) {
+                    context.ability.selectAnotherTarget(context.player, context, {
+
+                        activePromptTitle: 'Choose opposing dude to bedazzle',
+                        waitingPromptTitle: 'Waiting for opponent to select dude',
```
mmeldo, line 19: Since we are adding `source: this`, `waitingPromptTitle` is not needed as it will be constructed from the `source`.

mmeldo (changes requested):

server/game/cards/12-2T2D/Bedazzle.js
PesiTheWizard: Are the seperate messages fine as long as the user is at the start? I was never happy with my attempts to combine them.

server/game/cards/12-2T2D/Bedazzle.js
PesiTheWizard: It took a while parsing dudecard vs drawcard vs basecard, but ok.

server/game/cards/12-2T2D/Bedazzle.js
mmeldo: yeah, it is fine. I am also trying to combine them sometimes, but usually am not happy.

mmeldo (approved):

PR #1176, opened 2021-10-29
ignore CORS checks in dev mode.

stopfstedt:
this gets around any CORS checks when developing.

alternatively, we could hard wire in  a whitelist of domains for dev mode, like so:

```js
if (process.env.NODE_ENV === 'development') {
    this.io.set('origins', 'http://localhost:8080');
}
```

but that may be making too many assumptions about the given dev environment. 

as a follow-up item, the `origins` setting may be best extracted out into application config/env  vars or the likes. no magic strings!


mmeldo (approved):
Did not have much experience with CORS, but based on what was in the link this is OK. Also, it is just for development, but as you say allowing `*:*` would be assuming too much.

mmeldo:
New issue: #1180

PR #1184, opened 2021-10-30
Update Upkeep discard prompt

### LINKED ISSUE #1183: Upkeep discard prompt for dudes
Update upkeep discard prompt from simple one where player has to discard manually to a select card prompt with proper validation.

mmeldo:
closes #1183


PesiTheWizard:
No raiseEvent? Poor Backroom Deals and E Gui.

mmeldo:
> No raiseEvent? Poor Backroom Deals and E Gui.

I am changing this actually because I am currently implementing E Gui. 
Both cards react on a card being discarded during upkeep, and that is why there is new option passed to discard: `isUpkeep`. That means those cards will react on `onCardDiscarded` event checking if it was discarded because of upkeep - has `option.isUpkeep`. No need for new event.

PR #1337, opened 2021-12-19
Card: Michael "The Badger" Dodge

### LINKED ISSUE #232: TLS - Michael "The Badger" Dodge

sholder:
I went through several iterations here before it was all working, so lmk if this is not the best way to do things. In particular:

* Should the needed pull be listed as a cost? Manually calling handlePull() in the body of the handler just gave me the value of the pulled card, I couldn't find any other way to check the suit.
* Is thenExecute() needed? It wasn't clear to exactly what it does, I added it while fixing a bug with the unbooted dudes calculation (but the bug was actually elsewhere, I'm pretty sure).

closes #232

### DIFF
--- a/server/game/cards/06-TLS/MichaelTheBadgerDodge.js
+++ b/server/game/cards/06-TLS/MichaelTheBadgerDodge.js
@@ -55,7 +55,7 @@ class MichaelTheBadgerDodge extends DudeCard {
                                     ability.effects.modifyBullets(-3)
                                 ]
                             }));
-                            this.game.addMessage('{0} uses {1} to lower the bullets of {2}', context.player, this, context.target);
+                            this.game.addMessage('{0} uses {1} to lower the bullets of {2} by 3', context.player, this, context.target);
                         }
                     }, true, { context });
                 });
--- a/server/game/cards/06-TLS/MichaelTheBadgerDodge.js
+++ b/server/game/cards/06-TLS/MichaelTheBadgerDodge.js
@@ -30,7 +30,7 @@ class MichaelTheBadgerDodge extends DudeCard {
                 },
                 cardType: ['dude']
             },
-            message: context => this.game.addMessage('{0} uses {1} to boot an opposing dude', context.player, this),
+            message: context => this.game.addMessage('{0} uses {1} to boot {2}', context.player, this, context.target),
             handler: context => {
                 this.game.resolveGameAction(GameActions.bootCard({ card: context.target }), context).thenExecute(() => {
                     const opposingUnbootedDudes = this.unbootedDudeCount(context.player.getOpponent());
--- a/server/game/cards/06-TLS/MichaelTheBadgerDodge.js
+++ b/server/game/cards/06-TLS/MichaelTheBadgerDodge.js
@@ -1,0 +1,75 @@
+const DudeCard = require('../../dudecard.js');
+const GameActions = require('../../GameActions/index.js');
+
+/** @typedef {import('../../AbilityDsl')} AbilityDsl */
+
+class MichaelTheBadgerDodge extends DudeCard {
+    /** @param {AbilityDsl} ability */
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            targetController: 'current',
+            condition: () => this.hasAttachmentWithKeywords(['melee', 'weapon']), 
+            match: this,
+            effect: [
+                ability.effects.setAsStud()
+            ]
+        });
+        this.action({
+            title: 'Michael "The Badger" Dodge',
+            playType: ['shootout'],
+            cost: [
+                ability.costs.bootSelf(),
+                // Not strictly a cost, but needed so we can reference the result in the 
+                // action handler.
+                ability.costs.pull()
+            ],
+            target: {
+                activePromptTitle: 'Choose an opposing Dude',
+                choosingPlayer: 'current',
+                cardCondition: { 
+                    inOpposingPosse: true, 
+                    controller: 'opponent',
+                    condition: card => !card.booted
+                },
+                cardType: ['dude'],
+                gameAction: 'boot'
+            },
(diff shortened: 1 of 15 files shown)


server/game/cards/06-TLS/MichaelTheBadgerDodge.js
```diff
@@ -0,0 +1,75 @@
...
+            ]
+        });
+        this.action({
+            title: 'Michael "The Badger" Dodge',
+            playType: ['shootout'],
+            cost: [
+                ability.costs.bootSelf(),
+                // Not strictly a cost, but needed so we can reference the result in the 
+                // action handler.
+                ability.costs.pull()
```
mmeldo, line 24: I am not really against it being a cost, but if you do not want it, you can implement it similar to for example `TheRDRanch.js`

server/game/cards/06-TLS/MichaelTheBadgerDodge.js
```diff
@@ -0,0 +1,75 @@
...
+                    const opposingUnbootedDudes = this.unbootedDudeCount(context.player.getOpponent());
+                    const unbootedDudes = this.unbootedDudeCount(context.player);
+                    if(opposingUnbootedDudes > unbootedDudes) {
+                        this.applyAbilityEffect(context.ability, ability => ({
+                            match: this,
+                            effect: [
+                                ability.effects.modifyBullets(-3)
+                            ]
+                        }));
+                        this.game.addMessage('{0}\'s bullets are lowered because there are more opposing unbooted dudes', this);
```
mmeldo, line 49: We have a rule that a message should start with a player.

server/game/cards/06-TLS/MichaelTheBadgerDodge.js
```diff
@@ -0,0 +1,75 @@
...
+                    if(context.pull.pulledSuit.toLowerCase() === 'clubs') {
+                        this.game.addMessage('{0} uses {1} but fails to lower the bullets of {2}', context.player, this, context.target);
+                    } else {
+                        context.target.applyAbilityEffect(context.ability, ability => ({
+                            match: context.target,
+                            effect: [
+                                ability.effects.modifyBullets(-3)
+                            ]
+                        }));
+                        this.game.addMessage('{0} uses {1} to lower the bullets of {2}', context.player, this, context.target);
```
mmeldo, line 61: If possible and it does not make the message too big, give as much in the message as you can.
```suggestion
                        this.game.addMessage('{0} uses {1} to lower the bullets of {2} by 3', context.player, this, context.target);
```

server/game/cards/06-TLS/MichaelTheBadgerDodge.js
```diff
@@ -0,0 +1,75 @@
...
+            target: {
+                activePromptTitle: 'Choose an opposing Dude',
+                choosingPlayer: 'current',
+                cardCondition: { 
+                    inOpposingPosse: true, 
+                    controller: 'opponent',
+                    condition: card => !card.booted
+                },
+                cardType: ['dude'],
+                gameAction: 'boot'
```
mmeldo, line 35: The `gameAction` in the target is used mostly to check if the target can be selected (meaning if that action can be performed on the target), and if subsequently the ability can be triggered.
This would be ok with the old rules as the first sentence was a requirement, so if you could not boot the target, you would not be able to trigger the ability.
However with new rules, it would not be required, as there is that "do as much as you can". So it would be possible to not boot it, but just give it -3 bullets based on the pull.

mmeldo (changes requested):

mmeldo:
`thenExecute` is used after an event, if you want to perform something only after the event has been executed. So if you want to perform something after you discard a card using `GameAction.discard` (and GameActions are an events), and even based it on a discarded card, you will use the `thenExecute(event => ...)`

server/game/drawcard.js
```diff
@@ -300,6 +300,9 @@ class DrawCard extends BaseCard {
     }
     
     getAttachmentsByKeywords(keywords) {
+        if(!this.attachments) {
```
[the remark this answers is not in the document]
sholder: Yeah, I was getting an error here during setup - maybe because I put The Badger in my starting posse for ease of testing? I don't recall the exact phase of the game now, but it was happening reliably on game start until I added this check.

server/game/cards/06-TLS/MichaelTheBadgerDodge.js
sholder: Done

server/game/cards/06-TLS/MichaelTheBadgerDodge.js
sholder: I like the flow better when its not a cost (the pull message occurs in a more natural order in the chat log) so I switched it. Thanks for the pointer.

server/game/cards/06-TLS/MichaelTheBadgerDodge.js
sholder: Forgot about the rule change, thanks!

server/game/cards/06-TLS/MichaelTheBadgerDodge.js
```diff
@@ -0,0 +1,73 @@
...
+                activePromptTitle: 'Choose an opposing Dude',
+                choosingPlayer: 'current',
+                cardCondition: { 
+                    inOpposingPosse: true, 
+                    controller: 'opponent',
+                    condition: card => !card.booted
+                },
+                cardType: ['dude']
+            },
+            message: context => this.game.addMessage('{0} uses {1} to boot an opposing dude', context.player, this),
```
mmeldo, line 33: The PR looks good, just one last change:
```suggestion
            message: context => this.game.addMessage('{0} uses {1} to boot {2}', context.player, this, context.target),
```

mmeldo (changes requested):

mmeldo (approved):

PR #1360, opened 2022-01-10
Fix currently failing unit test

sholder:
`npm test` current fails - the logic for Shotgun.spec.js was updated to use logic for the WWE edition of Shotgun, but the test is still picking the DTR version. Given the test file is in the "01-DTR" directory, I changed the logic back to match the DTR version of the car, lmk if you agree.


mmeldo:
This is already fixed in master by updating the test that checks when value = bullets.

sholder:
sg, abandoning this PR then

mmeldo:
the test is failing because you do not have the `townsquare-json-data` submodule updated in your `townsquare` project.

Issue #199, opened 2021-03-23
Incorrect example in the Implementing Cards document


mmeldo:
Submitter: Aaron Schnabel
To fix in the contribution guide
- [x] In the Implementing Cards document, the example for aceCard is given under discardCard
- [x] Broken link to General information
- [x] Update examples due to the change for `message` property
- [x] Other typos and errors

PR #1370, opened 2022-01-14
Minor documentation cleanup

sholder:
This should mostly resolve issue #199, though I couldn't figure out what was required for "Update examples due to the change for message property". Docs and code in that area seemed to be in sync from what I could tell.


docs/implementing-cards.md
```diff
@@ -407,7 +417,7 @@ this.applyAbilityEffect(context.ability, ability => ({
 }));
 ```
 
-In case the duration of effects was changes (e.g. Orphanage in third example), use following functions.
+In case the duration of effects was changes, use following functions.
```
mmeldo, line 420:
```suggestion
In case the duration of effects was changed, use following functions.
```

docs/implementing-cards.md
```diff
@@ -1023,17 +1034,18 @@ handler: card => {
...
 
 **bootCard** - can be called using `player.bootCard(card, playType)`. Use playType parameter only if you are performing special kind of booting.
+
 **callOut**
  - `caller`: dude who is doing the Call out
  - `callee`: dude who is being Called out
- - `isCardEffect` (true): *true* if the call out comes from the effect of a card.
- - `canReject` (true): *true* if the callee can reject the call out.
+ - `isCardEffect`: (true): *true* if the call out comes from the effect of a card.
+ - `canReject`: (true): *true* if the callee can reject the call out.
```
mmeldo, line 1042: I believe this was correct as `(true)` was meant as default.
```suggestion
 - `isCardEffect` (true): *true* if the call out comes from the effect of a card.
 - `canReject` (true): *true* if the callee can reject the call out.
```

mmeldo (commented):
Thanks a lot for this :) I did just a few small comments.

mmeldo (approved):

PR #1383, opened 2022-01-22
Update NullCard with isNearby()

sholder:
Add default implementation of isNearby() to NullCard since some cards will attempt to call it.

### DIFF
--- a/server/game/nullcard.js
+++ b/server/game/nullcard.js
@@ -34,6 +34,10 @@ class NullCard {
     adjacentLocations() {
         return [];
     }
+
+    isNearby() {
+        return false;
+    }
 }
 
 module.exports = NullCard;


mmeldo (approved):
Yeah, that should do it. And it looks like I should finally properly refactor the card class hierarchy to avoid these problems.

sholder:
> Yeah, that should do it. And it looks like I should finally properly refactor the card class hierarchy to avoid these problems.

Happy to help out if it would be useful, or even just contribute unit test coverage. This was my first time digging into this part of the code, I'll poke around a bit and try to get familiar with the current setup.

mmeldo:
> Happy to help out if it would be useful, or even just contribute unit test coverage. This was my first time digging into this part of the code, I'll poke around a bit and try to get familiar with the current setup.

Oh man, if you would help with the unit test coverage, you would be my hero 😄 That is something I try to force myself to do for quite some time because there is a lot of things still not covered.
There are few issues I created, but that is not all: https://github.com/townteki/townsquare/issues?q=is%3Aissue+is%3Aopen+label%3A%22tests+%3Asafety_vest%3A%22

Or maybe grab a refactor issue: https://github.com/townteki/townsquare/issues?q=is%3Aissue+is%3Aopen+label%3A%22refactor+%3Abomb%3A%22

or a bug. It is up to you, I will be happy with any help really :)
Of course if you have any questions, just ping me on discord.

PR #1385, opened 2022-01-23
Get code coverage working.

sholder:
The previous coverage command was erroring out for me, and the package it depends on (Istanbul) seems to be no longer maintained. This installs `nyc` (https://www.npmjs.com/package/nyc) and configures the project so that `npm run coverage` produces a coverage report.

The current configuration **does** include all individual cards in the coverage report, many of which have little or no coverage. lmk if you would rather exclude them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ testoutput/*
 dist/*
 .vscode/tasks.json
 local.json5
+.nyc_output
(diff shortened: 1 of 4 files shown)


mmeldo:
To be honest, I am not sure if to include or not. But they should be covered as well, with a specific cards having their own tests.

mmeldo (approved):

PR #1392, opened 2022-02-05
Card: GT - Framed

PesiTheWizard:
* Adding file for Framed

* Marking Framed

_(It's been a while, I hope I haven't forgotten something basic.
I added the card-title to the prompt-title because Gomorra Gaming Commission has shown that players don't know what is happening when the prompt first shows up)_


mmeldo:
> I added the card-title to the prompt-title because Gomorra Gaming Commission has shown that players don't know what is happening when the prompt first shows up

There is no need to add `Framed` to the prompt title if you have `source: this`. That will cause that the source card title will appear as a prompt title (text on top of prompt separated by a line). I see that there is no `source: this` in `Gomorra Gaming Commission` so that is why the card title is missing.

mmeldo (approved):

PR #1400, opened 2022-02-18
Card: GT - Sunday Best

PesiTheWizard:
* Adding file for Sunday Best

* Marking Sunday Best


mmeldo (changes requested):

mmeldo:
Did not merge this one because the build is failing.

PesiTheWizard:
So what was with the apostrophe-change anyway?

PesiTheWizard:
This seems to just be made of fail. And I apparently forgot the boot cost. I think it's back to the drawing board.

mmeldo:
> So what was with the apostrophe-change anyway?

with the normal apostrophes (' or "), everything in between them is considered a regular text.
but if you use \`, you can do a variable substitution by using `${variable}` in the text.

PR #1406, opened 2022-02-22
Some fixes

mmeldo:
Fix Ramiro Mendoza to ask to pay 1 GR, timer start and cards with upkeep other than dudes.


PesiTheWizard:
This would still mean non-dudes can't be fired.

https://github.com/townteki/townsquare/blob/250cb5e0c8cc92110be54958d330a8899ab5066c/server/game/gamesteps/upkeep/upkeepprompt.js#L107

mmeldo:
I thought you can fire only dudes, but you are right, it can be any card with upkeep.

PR #1412, opened 2022-03-13
Misc fixes 2

PesiTheWizard:
* Removing unfinished message

Full message is already in changeBountyLaura()

* Fixing ref count

It counted 0, 2, 3

* Typo

looses -> loses

* PlayType fix

Ability was unusable

* Update handresult.js

Heretic joker can beat up Devils joker (If both present, cheatness depended on which joker was rightmost in the hand)

* Fix Type Error in case shootout gets into weird state

* Fix Putting the Pieces Together to not require action cost when it is in play

part of #1412


mmeldo (approved):

PesiTheWizard:
Somehow the playType change of ragged mule went from this PR to #1397

PR #1420, opened 2022-03-19
Card: DTR - Max Baine

PesiTheWizard:
* Adding file for Max Baine

* Marking Max Baine

### DIFF
--- a/server/game/cards/01-DTR/MaxBaine.js
+++ b/server/game/cards/01-DTR/MaxBaine.js
@@ -1,0 +1,28 @@
+const DudeCard = require('../../dudecard.js');
+
+class MaxBaine extends DudeCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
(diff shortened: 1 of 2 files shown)


server/game/cards/01-DTR/MaxBaine.js
```diff
@@ -0,0 +1,28 @@
...
+                if(card.hasKeyword('ranch')) {
+                    return count + 1;
+                }
+                return count;
+            }, 0) >= 2,
+            match: this,
+            effect: ability.effects.modifyControl(1)
+        });
+        this.persistentEffect({
+            targetController: 'current',
```
mmeldo, line 16: State dependent effects need to have a condition, otherwise they are not reevaluated.
```suggestion
            targetController: 'current',
            condition: () => true,
```

mmeldo (changes requested):

mmeldo (approved):

PR #1431, opened 2022-06-06
Random Fixes 3

PesiTheWizard:
* TheFlyingPopescus.js

Somehow isn't getting the infuence bonus. The hope is that arrayifying the getAttachmentsByKeyword input (to match every other use of that function) will fix this.

* HighStakesHaven.js

Having no cards in hand brought up the prompt where discard could be clicked, even when GR was available.
Total rework of handler.
Added false to discardAtRandom-call to prevent double-logging

* BackWays.js

PromptForLocation should return true; it was my mistake.
Location prompt stuck around and you could send the dude all over until finally clicking cancel.

* QuickdrawHandgun2.js

1) The user of the gun is the one who takes the illegal hand, and so the one who needs legalization.
2) As .determineHandResult() creates a new .handResult and subsequently a new .handRank, fiddling with the .handRank before calling .determineHandResult() is pointless.
The hand-changing message will still say the new hand is cheatin', but that now gets fixed before anything relevant can happen.

* RailroadStation.js

Prompt title includes target's name; destination must be different.

* GomorraGamingCommission.js

Finally adding source:this
Hopefully users will no longer be confused why there's suddenly a prompt

* InnerStruggle.js

Adding special message case for handless cheater.
There used to be no message in this case.

* Tusk

In case 2 Tusks were in play and opponent used a card ability, the prompt for trait order was displayed with 2 Tusk buttons.

### DIFF
--- a/server/game/cards/09.1-TCR/HighStakesHaven.js
+++ b/server/game/cards/09.1-TCR/HighStakesHaven.js
@@ -7,19 +7,38 @@ class HighStakesHaven extends DeedCard {
                 onDrawHandsRevealed: () => this.controller.isCheatin()
             },
             handler: () => {
-                if(this.controller.getSpendableGhostRock() <= 0) {
-                    this.havenRandomDiscard();
-                } else if(this.controller.hand.length) {
-                    this.game.promptWithMenu(this.controller, this, {
-                        activePrompt: {
-                            menuTitle: 'Choose cheatin\' punishment',
-                            buttons: [
-                                { text: 'Pay 1 GR', method: 'payGR' },
-                                { text: 'Discard card', method: 'havenRandomDiscard' }
-                            ]
-                        },
-                        source: this
-                    });
+                let choicemask = 0;
+                if(this.controller.getSpendableGhostRock() >= 1) {
+                    choicemask = choicemask | 1;
+                }
+                if(this.controller.hand.length) {
+                    choicemask = choicemask | 2;
+                }
+                switch(choicemask) {
+                    case 0://has no GR and no cards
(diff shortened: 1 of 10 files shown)


server/game/cards/09.2-AGE/InnerStruggle.js
```diff
@@ -30,10 +30,14 @@ class InnerStruggle extends ActionCard {
                 const theirhome = context.player.getOpponent().getOutfitCard();
                 context.player.attach(this, theirhome, 'ability', () => {
                     this.booted = true;
-                    this.controller.discardAtRandom(1, discarded => {
-                        this.game.addMessage('{0} uses {1}, attaches it to {2} and {3} randomly discards {4}', 
-                            context.player, this, theirhome, this.controller, discarded);
-                    }, false);
+                    if(this.controller.hand.length) {
```
mmeldo, line 33: I would rather fix this in the `discardAtRandom` function so the hand does not have to be checked every time.

server/game/cards/09.1-TCR/HighStakesHaven.js
```diff
@@ -7,19 +7,38 @@ class HighStakesHaven extends DeedCard {
...
-                        activePrompt: {
-                            menuTitle: 'Choose cheatin\' punishment',
-                            buttons: [
-                                { text: 'Pay 1 GR', method: 'payGR' },
-                                { text: 'Discard card', method: 'havenRandomDiscard' }
-                            ]
-                        },
-                        source: this
-                    });
+                let choicemask = 0;
```
mmeldo, line 10: Similar to the `InnerStruggle`, I would rather change `discardAtRandom` function to return if the card was discarded.

mmeldo (approved):

PR #1443, opened 2022-07-14
Fixed so that player doesn't take casualties.

### LINKED ISSUE #1424: Disarm Error
**Card**
Disarm

**Describe the bug**
I used Disarm as a cheatin' res but was still prompted to take casualties.

**To Reproduce**
1. Do action '...'
3. Start shootout '....'

**Screenshots**

**Additional context**

AaronSchnabel:
closes #1424 . Fixed issue with casualty prevention being in the wrong if statement, and removed text about ignoring casualties when not in a shootout.


mmeldo:
I am little bit confused here. I get that the `decreaseCasualties` is in incorrect place, but how does it solve the problem? Before, it was in a place where it was triggered even if not in shootout, meaning it was triggered every time the card was used. But the bug is the opposite, that the casualties are not reduced at all.

mmeldo:
Should be fixed. 
@stopfstedt could you please review and test? thanks

stopfstedt (approved):
code reviewed and click tested. LGTM.

PR #1444, opened 2022-07-14
Fixed to allow booted dudes to join

### LINKED ISSUE #642: Buffalo Rifle won't allow booted dudes to join shootout.
**Card**
Buffalo Rifle

**Describe the bug**
Won't allow booted dude at adjacent location to join posse, even though card text allows for it. Unbooted dudes can join, no problem. The "sniper" dialog appears and works.

AaronSchnabel:
closes #642


stopfstedt (commented):
this card seems to be working fine as-is. The version from the original base set is the one that's buggy.

PR #1445, opened 2022-07-14
Added town square as a legal location

### LINKED ISSUE #1433: Ghostly Communion
**Card**
Ghostly Communion

**Describe the bug**
Tried to use Ghostly Communion while my dude was at an OOT deed.  It did not allow me to move them to TS.  Because my home is seen as a Holy Ground while casting the spell it should be allowed. 

**To Reproduce**
1. Do action '...'
3. Start shootout '....'

**Screenshots**

**Additional context**

AaronSchnabel:
closes #1433


server/game/cards/16-DT2/GhostlyCommunion2.js
```diff
@@ -15,7 +15,7 @@ class GhostlyCommunion2 extends SpellCard {
                     location: 'play area',
                     condition: card => card.hasKeyword('holy ground') || 
                         card.adjacentLocations().some(location => location.locationCard.hasKeyword('holy ground')) ||
-                        card.isNearby(this.controller.outfit.uuid)
+                        card.isNearby(this.controller.outfit.uuid) || this.game.townsquare.uuid
```
stopfstedt, line 18: this change would always result in all adjacent locations of every holy ground location to become legal targets. since `this.game.townsquare.uuid` is `townsquare`, which is a truthy value so this condition would always return OK for any card given.

to check if the given card represents the townsquare, you could do a comparison like that instead:

`|| this.game.townsquare.uuid === card.uuid`

stopfstedt (commented):
~~as commented on in the linked issue, this appears to be working as-is/no changes required.~~

scratch that, i was testing with an outfit that is "Holy Ground" :man_facepalming:

server/game/cards/16-DT2/GhostlyCommunion2.js
```diff
@@ -37,7 +37,7 @@ class GhostlyCommunion2 extends SpellCard {
             condition: () => this.game.shootout &&
                 (this.game.shootout.shootoutLocation.locationCard.hasKeyword('holy ground') || 
                 this.game.shootout.shootoutLocation.locationCard.adjacentLocations().some(location => location.locationCard.hasKeyword('holy ground')) ||
-                this.game.shootout.shootoutLocation.locationCard.isNearby(this.controller.outfit.uuid)),
+                this.game.shootout.shootoutLocation.locationCard.isNearby(this.controller.outfit.uuid)) || this.game.shootout.shootoutLocation === this.game.townsquare.uuid,
```
stopfstedt, line 40: the shootout location's location card has to be used for comparison. 
and the conditional should be inside the parenthesis with the other  shootout location checks.

```suggestion
                this.game.shootout.shootoutLocation.locationCard.isNearby(this.controller.outfit.uuid) ||
                this.game.shootout.shootoutLocation.locationCard.uuid === this.game.townsquare.uuid
                ),
```

stopfstedt (changes requested):
this isn't quite correct yet.
please see suggested changes.

server/game/cards/16-DT2/GhostlyCommunion2.js
mmeldo: You do not need to get the `locationCard` of the `shootoutLocation`. The class `GameLocation` (class of shootoutLocation) has an `uuid` getter, so you can just do this:
```
this.game.shootout.shootoutLocation.uuid === this.game.townsquare.uuid
```

mmeldo:
Fixed it in a way that will prevent this in other cards. But also I realized that the town square would really need its own card.
@stopfstedt could you please review and test? thanks

stopfstedt (approved):
click tested, LGTM.

PR #1450, opened 2022-07-22
Card: EDS - Teleportation Device

PesiTheWizard:
* Adding file for Teleportation Device

* Marking Teleportation Device

### DIFF
--- a/server/game/cards/02.3-EDS/TeleportationDevice.js
+++ b/server/game/cards/02.3-EDS/TeleportationDevice.js
@@ -1,0 +1,60 @@
+const GoodsCard = require('../../goodscard.js');
+const GameActions = require('../../GameActions/index.js');
+
+class TeleportationDevice extends GoodsCard {
+    setupCardAbilities(ability) {
+        this.attachmentRestriction(card => 
+            card.controller === this.controller &&
+            card.getType() === 'deed'
+        );
+
+        this.action({
+            title: 'Noon: Teleportation Device',
+            playType: ['noon'],
+            repeatable: true,
+            cost: [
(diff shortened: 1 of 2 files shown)


server/game/cards/02.3-EDS/TeleportationDevice.js
```diff
@@ -0,0 +1,60 @@
...
+                    });
+                } else {
+                    this.game.resolveGameAction(GameActions.moveDude({ 
+                        card: context.targets.tDude, 
+                        targetUuid: context.targets.tDest.uuid 
+                    }), context).thenExecute(() => {
+                        teleMessage += ', moving them to {3}';
+                    });
+                }
+                this.game.addMessage(teleMessage, context.player, this, context.targets.tDude, context.targets.tDest);
```
mmeldo, line 52: The `teleMessage` will always be the default one ('{0} uses {1} on {2}') because the `addMessage` will be performed before the code in `thenExecute`.

mmeldo (changes requested):

server/game/cards/02.3-EDS/TeleportationDevice.js
mmeldo: I do not like this solution, it is too complicated and the allow check will be done twice. As you mentioned, you can just use `queueSimpleStep`:
```javascript
                this.game.queueSimpleStep(() => {                 
                    this.game.addMessage(teleMessage, context.player, this, context.targets.tDude, context.targets.tDest);
                });
```

mmeldo (changes requested):

mmeldo (approved):

PR #1465, opened 2022-08-09
Card: FP - Martyr's Cry

PesiTheWizard:
* Adding file for Martyr's Cry

* Marking Martyr's Cry

### DIFF
--- a/server/game/cards/07.2-FP/MartyrsCry.js
+++ b/server/game/cards/07.2-FP/MartyrsCry.js
@@ -1,0 +1,29 @@
+const ActionCard = require('../../actioncard.js');
+const GameActions = require('../../GameActions/index.js');
+
+class MartyrsCry extends ActionCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Resolution: Martyr\'s Cry',
+            playType: 'resolution',
+            cost: ability.costs.discardFromPlay({
+                type: 'spell',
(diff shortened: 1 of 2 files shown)


mmeldo:
Again, thanks a lot for the card implementation, but it would help me immensely if you also tested it in your dev environment. Due to my current time constraints, I will have to put it at the back of the queue.

server/game/cards/07.2-FP/MartyrsCry.js
```diff
@@ -0,0 +1,29 @@
...
+const GameActions = require('../../GameActions/index.js');
+
+class MartyrsCry extends ActionCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Resolution: Martyr\'s Cry',
+            playType: 'resolution',
+            cost: ability.costs.discardFromPlay({
+                type: 'spell',
+                condition: card => card.isMiracle() && card.parent.isParticipating()
```
mmeldo, line 11: Also check if there is `card.parent` to be absolutely sure.

mmeldo (changes requested):

server/game/cards/07.2-FP/MartyrsCry.js
PesiTheWizard: Notably, this part came from Hex Slingin', which does not have that check.

mmeldo (approved):

PR #1466, opened 2022-08-09
gameAction addition patch

PesiTheWizard:
gameAction additions for YanLisTailoring.js; WarPaint.js; BBAttorneys.js; SunInYerEyes.js; Framed.js; CliffsNo4Saloon.js; EBFarnum.js

### DIFF
--- a/server/game/cards/01-DTR/SunInYerEyes.js
+++ b/server/game/cards/01-DTR/SunInYerEyes.js
@@ -8,8 +8,7 @@ class SunInYerEyes extends ActionCard {
             target: {
                 activePromptTitle: 'Select a dude who has sun in their eyes',
                 cardCondition: { location: 'play area', participating: true, controller: 'opponent' },
-                cardType: ['dude'],
-                gameAction: ['decreaseBullets', 'setAsDraw']
+                cardType: 'dude'
             },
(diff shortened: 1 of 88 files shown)


server/game/cards/01-DTR/SunInYerEyes.js
```diff
@@ -8,7 +8,8 @@ class SunInYerEyes extends ActionCard {
             target: {
                 activePromptTitle: 'Select a dude who has sun in their eyes',
                 cardCondition: { location: 'play area', participating: true, controller: 'opponent' },
-                cardType: 'dude'
+                cardType: ['dude'],
+                gameAction: ['decreaseBullets', 'setAsDraw']
```
mmeldo, line 12: Not sure about the others where there is just one action, but this is not correct as it would prevent the whole effect if there is only one of the actions prevented.

PesiTheWizard: Surprising; it was [The Mayor's Office](https://github.com/townteki/townsquare/blob/master/server/game/cards/02.3-EDS/TheMayorsOffice.js#L12) that convinced me it worked the other way.

mmeldo (approved):

PR #1502, opened 2022-09-15
Fix Jokers used by card abilities

### LINKED ISSUE #1464: Draw hand changers don't look at phases when resolving jokers
Any card that invokes .determineHandResult has predetermined joker behaviour

The clearest examples:
* Using Devil's Six Gun during lowball has it maximize the rank
* Bad Beat and Bottom Dealin' always minimize the rank regardless of phase or who owns the joker

(There might be deeper issues when several of these intersect)

mmeldo:
fixes #1464


mmeldo:
No idea why this has been closed and not merged. Probably by mistake.

PR #1512, opened 2022-09-26
Card: FP - Deliberate Infection

PesiTheWizard:
* Adding file for Deliberate Infection

* Marking Deliberate Infection

* Adding thisIfShootout


server/game/cards/07.2-FP/DeliberateInfection.js
```diff
@@ -0,0 +1,36 @@
...
+                ability.effects.modifyInfluence(-1)
+            ]
+        });
+        this.action({
+            title: 'Deliberate Infection',
+            playType: 'cheatin resolution',
+            condition: () => !this.parent,
+            target: {
+                choosingPlayer: 'thisIfShootout',
+                cardCondition: { location: 'play area', controller: 'opponent', participating: true },
```
mmeldo, line 19: I do not think we want to set `participating: true` since this can be played during Gamblin' phase. I guess better would be to use condition and require target to be participating only if in shootout.

mmeldo (changes requested):

server/game/cards/07.2-FP/DeliberateInfection.js
PesiTheWizard: It seems to work fine for coachwhip.

PesiTheWizard: AFAICT CardMatcher.js will match participating as true any time a shootout isn't happening

mmeldo (approved):

